PR #2, opened 2015-02-02
Fix quotes with syntax properties

### LINKED ISSUE #1: Emacs: '"' (double quote character) breaks syntax highlighting
Apparently, `rust-mode` thinks that `"` starts a string (that's why in the next line `{}` is not highlighted).

screen shot 2015-01-30 at 14 30 32

### LINKED ISSUE #6: emacs mode: ']', ')', and '}' close previous open paren/brace/bracket
<a href="https://github.com/pnkfelix"></img></a> **Issue by [pnkfelix](https://github.com/pnkfelix)**
_Monday Nov 04, 2013 at 15:42 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/10262_

----

Here is the illustrative example:

```rust
fn foo() {
    { bar('}'); }
    { bar(']'); }
    { bar(')'); }
}
```

The characters within `'` characters above are incorrectly treated by `src/etc/rust-mode.el` as closing the opening paren of the invocation of `bar`.  This confuses Emacs in a number of ways (e.g. obviously paren matching; but also the auto-indentation gets confused too, which is more annoying).

MicahChalmer:
This uses syntax properties to make it so that emacs recognizes the single quote, rather than the double quote, as the string delimiter within character literals, while leaving the syntax unchanged elsewhere.

Also included is a commit that restores ERT tests that were inadvertently broken by an earlier commit (removing trailing whitespace broke the tests--I restored them and made it so they are no longer dependent on having trailing whitespace in the code.)

Fixes #1.
Fixes #6.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -246,8 +246,8 @@ fn bar() { }" 14 67))
    "/**
  *
  */"
-   8
-   "This is a very very very very very very very long string"
+   7
+   " This is a very very very very very very very long string"
    "/**
  * This is a very very very very
  * very very very long string
@@ -317,8 +317,7 @@ fn foo() {
     /*!
      * this is a nested doc comment
      */
-
-    //! And so is this
+    \n    //! And so is this
 }"))
 
 (ert-deftest indent-inside-braces ()
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -899,3 +899,11 @@ list of substrings of `STR' each followed by its face."
    "'\"'; let"
    '("'\"'" font-lock-string-face
      "let" font-lock-keyword-face)))
+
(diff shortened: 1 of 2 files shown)


pnkfelix:
Are there any tests of character literal parsing?

I would in particular like to see tests of `'\''` (i.e. the single quote character); (I am inferring that is handled by this code, but I haven't read it quite carefully enough to confirm that.)

MicahChalmer:
I added a test for `'\''`.  I also confirmed that this same change also fixes #6, and added a test for that.  (I updated the pull request body to reflect that.)

pnkfelix:
@MicahChalmer hey thanks for following up; I took the liberty of cherry-picking the first commit in the series because that seemed important for overall workflow (at least for me).  I'll be looking at the other commits shortly.

PR #16, opened 2015-02-04
Add travis config (and make tests runnable on emacs 23)

MicahChalmer:
There was a bit more to this than I initially expected...

The rust-mode code has a few things in it that show intent to keep it compatible with emacs 23 as well as emacs 24, so the `.travis.yml` file here tests it against emacs 23, 24 and the latest snapshot build.  To make it successfully test on emacs 23 I had to change the test code (though not the actual rust-mode code under test) to eliminate some use of forms that were newly introduced in emacs 24.

The way it installs emacs in the Travis VM is taken from the example at https://github.com/rolandwalker/emacs-travis.  Other repos seem using the same approach as well--a very well known one is [Magit](https://github.com/magit/magit/blob/master/.travis.yml).  

There are a few unfortunate things about it:

  1. The PPA that it uses to get emacs 24 [now proclaims itself deprecated](https://launchpad.net/~cassou/+archive/ubuntu/emacs), although for the time being it still works.
  2. Because it uses apt-get to install emacs, it can't use Travis's new container infrastructure.  The Travis team appears to be working on lifting this limitation, so this seems to be only a temporary problem.

An alternative approach would be to have it download an emacs source tarball right from a GNU mirror, and rebuild emacs from source each time.  That would eliminate dependency on no-longer-interested third parties, but would also complicate the configuration in this repo and might make the travis builds take a while.  (I haven't tried it, so I can't really say for sure.)

Here is [a successful Travis run with this version from my own fork](https://travis-ci.org/MicahChalmer/rust-mode/builds/49408534).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,41 @@
+language: generic
+
+env:
+  matrix:
+    - EMACS=emacs23
+    - EMACS=emacs24
+    - EMACS=emacs-snapshot
+
+install:
+  - if [ "$EMACS" = 'emacs23' ]; then
+       sudo apt-get -qq update &&
+       sudo apt-get -qq -f install &&
+       sudo apt-get -qq install emacs23-gtk emacs23-el;
+    fi
+  - if [ "$EMACS" = 'emacs24' ]; then
+       sudo add-apt-repository -y ppa:cassou/emacs &&
+       sudo apt-get -qq update &&
+       sudo apt-get -qq -f install &&
+       sudo apt-get -qq install emacs24 emacs24-el;
+    fi
+  - if [ "$EMACS" = 'emacs-snapshot' ]; then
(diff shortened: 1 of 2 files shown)


nikomatsakis:
This looks pretty reasonable, and it'd be *awesome* to have travis running the unit tests.

nikomatsakis:
@MicahChalmer what else is needed to get travis going?

MicahChalmer:
All it needs now is for someone with admin access on this repo to go to travis-ci.org and flip the switch.

huonw:
Done.

huonw:
Looks like it works great, thanks @MicahChalmer: https://github.com/rust-lang/rust-mode/pull/24, https://travis-ci.org/rust-lang/rust-mode/builds/49740310 (tests seem to fail on emac23?).

PR #22, opened 2015-02-06
Fix aligning of method chains (more-or-less), add various unit tests, and add matching angle brackets.

### LINKED ISSUE #13: Emacs: rust-indent-method-chains indents beginning of line oddly
<a href="https://github.com/ProjectMoon"></img></a> **Issue by [ProjectMoon](https://github.com/ProjectMoon)**
_Monday Jan 26, 2015 at 10:29 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/21650_

----

Trying out rust-mode in emacs, and set rust-indent-method-chains because I like me some aligned dots.

It does what it advertises, but for some reason shoves the beginning of the line (and thus the lines underneath it) over one more tab.

I'm using spaces for tabs, configured to width 4.

What I expect:

```rust
let x = thing.do_it()
             .aligned()
             .more_alignment();
```

What I get:

```rust
    let x = thing.do_it()
                 .aligned()
                 .more_alignment();
```

I am using el-get to install rust-mode so it should be pulling off the latest master.

### LINKED ISSUE #23: Highlight matching angle brackets
We have some structs with very long nested types:

```rust
type SqrtCli<R, S, T> =
    Choose<Eps,
    Choose<R,
    Choose<S,
    Choose<Send<f64, Offer<Recv<f64, Var<Z>>, Var<Z>>>,
    T>>>>;
```

It would be very nice to have the syntax highlightingin rust-mode work for angle brackets as well --- as in, highlighting matching delimiters ---  since it is a pain to keep track of them yourself. This goes for type parameters in functions, structures, impls and traits as well.

I realize that the problem is to parse `<` and `>` differently whether they're used as delimiters or simple less-than/greater-than signs --- a problem that the rustc parser has as well --- but this is something that would help us tremendously. I'm not very familiar with writing indentation modes but if there's any way I can help, please let me know.

It'd probably also be desirable for rust-mode to be able to indent type parameters properly, but the above code isn't the best example of how to do that.

nikomatsakis:
- Fixes indentation for method chains (which I suspect we should default to "on").
- Causes matching `<` to be briefly highlighted when `>` is typed (configurable to "off", defaults to "on").

Caveat for method chains: The regular expressions are probably kind of wrong for things like

    self.foo.something::<X>()
            .more();

Fixes #13.
Fixes #23.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -919,3 +919,89 @@ list of substrings of `STR' each followed by its face."
      "main" font-lock-function-name-face
      "let" font-lock-keyword-face
      "'\\''" font-lock-string-face)))
+
+(ert-deftest indent-method-chains-no-align ()
+  (let ((rust-indent-method-chain nil)) (test-indent
+   "
+fn main() {
+    let x = thing.do_it()
+        .aligned()
+        .more_alignment();
+}
+"
+   )))
+
+(ert-deftest indent-method-chains-with-align ()
+  (let ((rust-indent-method-chain t)) (test-indent
+   "
+fn main() {
+    let x = thing.do_it()
+                 .aligned()
+                 .more_alignment();
+}
+"
+   )))
+
+(ert-deftest indent-method-chains-with-align-and-second-stmt ()
+  (let ((rust-indent-method-chain t)) (test-indent
+   "
+fn main() {
+    let x = thing.do_it()
+                 .aligned()
+                 .more_alignment();
+    foo.bar();
+}
+"
+   )))
+
+(ert-deftest indent-method-chains-field ()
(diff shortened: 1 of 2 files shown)


pnkfelix:
Sweet!

(The angle-bracket matching is not quite smart enough to e.g. skip over the content of comments or string literals.  But I do not see that as a reason to reject it.)

nikomatsakis:
@pnkfelix yes, good point, it also doesn't detect mismatched parens that would be useful. *But* the presence of comments or string literals *within types* seems pretty unlikely in any case.

nikomatsakis:
(in any case a mismatch is fairly harmless)

PR #35, opened 2015-02-10
Fix raw string bugs

### LINKED ISSUE #34: End of string match failures
Using `rust-mode.el` 8d99bf8, the following highlights everything after "xs" as a string literal.  Editting either string causes the syntax highlighting to work correctly (and reloading the buffer breaks it again).

```
fn g() -> &str {
    "Er"
}

fn g() -> bool {
    assert!(false,"xs");
    true
}
```

MicahChalmer:
Fix #34, as well as some other odd cases that could fool the previously landed raw string handling (look in the tests for the details of the edge cases).

While this may look like a lot more code, it's mostly because I've expanded the regexp using the built-in `rx` sexp notation, and commented each piece so that hopefully it can be understood later.  It certainly made it easier for _me_ to work with, especially after I had to make a single regexp cover all raw strings--both with and without `#` characters--to avoid some problems that popped up if I kept it split into two.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -963,6 +963,54 @@ list of substrings of `STR' each followed by its face."
      "r\"With a backslash at the end\\\"" font-lock-string-face
      "r##\"With two hashes\"##" font-lock-string-face)))
 
+(ert-deftest font-lock-raw-string-with-inner-hash ()
+  (rust-test-font-lock
+   "r##\"I've got an octothorpe (#)\"##; foo()"
+   '("r##\"I've got an octothorpe (#)\"##" font-lock-string-face)))
+
+(ert-deftest font-lock-string-ending-with-r-not-raw-string ()
+  (rust-test-font-lock
+   "fn f() {
+    \"Er\";
+}
+
+fn g() {
+    \"xs\";
+}"
+   '("fn" font-lock-keyword-face
+     "f" font-lock-function-name-face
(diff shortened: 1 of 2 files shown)


rust-mode-tests.el
```diff
@@ -963,6 +963,54 @@ list of substrings of `STR' each followed by its face."
...
+   "r\"With what looks like the start of a raw string at the end r#\";
+not_a_string();
+r##\"With \"embedded\" quote \"##;"
+   '("r\"With what looks like the start of a raw string at the end r#\"" font-lock-string-face
+     "r##\"With \"embedded\" quote \"##" font-lock-string-face)))
+
+(ert-deftest font-lock-raw-string-starter-inside-raw-string ()
+  ;; Check that it won't look for a raw string beginning inside another raw string.
+  (rust-test-font-lock
+   "r#\"In the first string r\" in the first string \"#;
```
huonw: Incidentally, does `r##"string "# still same string"##` work?

MicahChalmer: With this change, yes it does.  Without this change it's broken.

huonw: Could you add a test for that form (that is, with an embedded quote-and-#)?

MicahChalmer: Sure...added as new commit below.

huonw:
Thanks.

PR #30, opened 2015-02-07
Emacs 23 Fixups

MicahChalmer:
Fix #27 as follows:

  * Fix the character quote handling to work on emacs 23
  * Fix the method chains to work on emacs 23
  * Update the ERT section of the README to show how to get ERT for emacs 23 (and to reflect #21)
  * Re-enable emacs 23 on travis, now that all the tests pass again

Also fixes #28, which was present in all emacs versions (I happened on it while fixing doing the rest of this, then saw that it had been filed as #28 in the meantime.)

PR #38, opened 2015-02-15
Fix emacs state corruption from incorrect "syntax-begin-function" setting

### LINKED ISSUE #36: impl fn indented to left margin
Under some circumstances, `impl` functions, such as `method_allows` below, are indented to the left margin.  I have tried to reduce this example further but have failed to reproduce the behaviour with anything simpler.

In some case indentation is also not fixed - invoking `indent-for-tab-command` several times results in toggling of indentation levels.

```
#[derive(RustcEncodable)]
struct Unit {
    name: String,
}
/// Handler for Units
#[derive(Debug,Clone)]
pub struct Units;
resource_handler!(Units);

impl Resource for Units {
    fn available_content_types(&self, _: &Request, _: &mut Response) -> Vec<Mime> {
        vec![Mime(TopLevel::Application, SubLevel::Json, vec![])]
    }

fn method_allowed(&self, req: &Request, _: &mut Response) -> Vec<Mime> {
        match req.method {
        method::Get|method::Head|method::Post => true,
        _ => false
        }
    }

    fn handle_ok(&self, _: &Request, resp: &mut Response) -> ResourceResult {
        "connection failed"
    }
}
```

MicahChalmer:
This is to fix #36.

Commit 6b57bbf set `syntax-begin-function` to `beginning-of-defun`.  This violated the contract for that function--it's supposed to get outside of all braces/parens--but functions could appear inside other blocks, so `beginning-of-defun` could end inside another pair of braces.  This in turn caused emacs corrupt some unknown internal state and cause itself to think there were fewer parens than there were.

The comments of the added test show how much I understood about the exact mechanism of the failure--which is not much.  I figure it's better to just fix it and leave the test there.  Though I don't understand the mechanism of the corruption, at least I understand how the code was violating the hook's contract.  Hopefully there isn't some other similar corruption hiding out that I haven't found yet.

@vhbit and @hugoduncan, does this fix the issues you were seeing?

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -293,7 +293,7 @@ very very very long string
        ;; The indentation will fail in some cases if the syntax properties are
        ;; not set.  This only happens when font-lock fontifies the buffer.
        (font-lock-fontify-buffer)
-       (indent-region 1 (buffer-size)))
+       (indent-region 1 (+ 1 (buffer-size))))
      indented)))
 
 
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1110,3 +1110,98 @@ fn main() { // comment here should not push next line out
 }
 "
    )))
+
+(ert-deftest test-for-issue-36-syntax-corrupted-state ()
+  "This is a test for a issue #36, which involved emacs's
+internal state getting corrupted when actions were done in a
+specific sequence.  The test seems arbitrary, and is, but it was
+not clear how to narrow it down further.
+
+The cause of the bug was code that used to set
+`syntax-begin-function' to `beginning-of-defun', which doesn't
+actually fulfill the expectations--`syntax-begin-function' is
+supposed to back out of all parens, but `beginning-of-defun'
+could leave it inside parens if a fn appears inside them.
+
+Having said that, as I write this I don't understand fully what
+internal state was corruped and how.  There wasn't an obvious
+pattern to what did and did not trip it."
+  
+  ;; When bug #36 was present, the following test would pass, but running it
+  ;; caused some unknown emacs state to be corrupted such that the following
+  ;; test failed.  Both the "blank_line" and "indented_closing_brace" functions
+  (with-temp-buffer
+    (rust-mode)
+    (insert "fn blank_line(arg:int) -> bool {
+
+}
+
+fn indenting_closing_brace() {
+    if(true) {
+}
+}
+
+fn indented_already() {
+    \n    // The previous line already has its spaces
+}
+")
+
+    (goto-line 11)
+    (move-to-column 0)
+    (indent-for-tab-command)
+    (should (equal (current-column) 4))
+    )
+
+  ;; This is the test that would fail only after running the previous one.  The
+  ;; code is extracted from src/libstd/collections/table.rs in the rust tree.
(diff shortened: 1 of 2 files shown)


rust-mode-tests.el
```diff
@@ -1110,3 +1110,98 @@ fn main() { // comment here should not push next line out
...
+not clear how to narrow it down further.
+
+The cause of the bug was code that used to set
+`syntax-begin-function' to `beginning-of-defun', which doesn't
+actually fulfill the expectations--`syntax-begin-function' is
+supposed to back out of all parens, but `beginning-of-defun'
+could leave it inside parens if a fn appears inside them.
+
+Having said that, as I write this I don't understand fully what
+internal state was corruped and how.  There wasn't an obvious
```
pnkfelix: typo: "corruped"

rust-mode-tests.el
```diff
@@ -1110,3 +1110,98 @@ fn main() { // comment here should not push next line out
...
+pattern to what did and did not trip it."
+  
+  ;; When bug #36 was present, the following test would pass, but running it
+  ;; caused some unknown emacs state to be corrupted such that the following
+  ;; test failed.  Both the "blank_line" and "indented_closing_brace" functions
```
pnkfelix: ... did this sentence have ....  a verb?

pnkfelix:
I have nits, but the fix seems as sound as far I understand it.  So I'm going to land this PR, and file an issue asking @MicahChalmer to fix the comments.  :)

pnkfelix:
Oh, one more question: @MicahChalmer can you elaborate (or point to an elaboration elsewhere) on *why* #30 included that setq of `syntax-begin-function` to `beginning-of-defun` ?  I.e. do you remember if there was a specific bug that you were addressing by doing that?

MicahChalmer:
I filed #40 to fix the typo and missing end of the comment.

As to @pnkfelix's  question of why I made the mistake of setting `syntax-begin-function` in #30 in the first place...when I wrote #30 I was trying to translate from the previous approach of using `syntax-propertize-function` and `syntax-propertize-extend-region`.  I was under the mistaken impression that `syntax-begin-function` was used in a similar way to the first element of the return value `syntax-propertize-extend-region`--that is, I thought it was used to start _further back_ and thus avoid starting intepretation in the middle of a construct.  But it's actually used to [optimize](http://www.gnu.org/software/emacs/manual/html_node/elisp/Position-Parse.html)--if `syntax-begin-function` is defined, it's only used to start _further forward_ than the default, rather than further back.  It's only there to make it go faster.  (There's a check in `syntax.el` that ignores the return value if it's further back than the default.)  Unless there are performance problems (and nobody has reported any so far) there is no reason to set it at all.

Of course, since I was also the one who put in the `syntax-propertize-extend-region` implementation as well, this naturally leads to follow-up questions of whether _that_ was necessary (and why I thought it was.)  I don't remember the answers to these.

I didn't put the above in the comments in #40--I figured it's a bit more than needed for ongoing maintenance of the code.  Let me know if you think it should go in anyway, and I can add it as another commit to #40.

vhbit:
@MicahChalmer I haven't chance to work a lot on Rust today, but on the quick look everything was fine, thanks!

hugoduncan:
It seems to have fixed things for me.  Many thanks for tracking that down.

Issue #41, opened 2015-02-18
{forward,backward}-word behavior is unexpected


roysc:
Because `_` is treated as a "word" character, functions operating on words do not behave in the usual way - eg. instead of jumping over "var" in  `var_name`, `forward-word` jumps over the whole symbol (ie. does the same thing as `forward-symbol`).
I tried to address this before but did not fix it properly, I will have a PR soon that is a proper fix.

MicahChalmer:
This seems to be the same issue as #11.  I would think either this one or #11 should be closed as a duplicate.

nikomatsakis:
Closing as dup of #11

PR #42, opened 2015-02-18
Fix word and symbol syntax distinction

### LINKED ISSUE #11: Emacs rust-mode: backward-word and backward-kill broken
<a href="https://github.com/fzzy"></img></a> **Issue by [fzzy](https://github.com/fzzy)**
_Thursday Nov 13, 2014 at 14:40 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/18923_

----

When I move backwards a word (M-b) or delete a word (M-d) while rust-mode and subword-mode are enabled, Emacs' text cursor will skip a snake cased subword.
I sometimes get similar problems with forward word moving as well, but not as often.
(If subword-mode is disabled, both M-f and M-b will advance whole snake cased words, which it doesn't normally do)

I'm not sure if this is just a issue with me, but this problem only occurs with rust-mode and not any other language modes, so I assume it's something with rust-mode.
I don't have any weird settings to my knowledge that should break this.

eg.

what i get
```
before M-b

pub fn mount(&self, path: &str, mount_point: &str, append_to_path: bool) {}
                                           ^

after M-b:

pub fn mount(&self, path: &str, mount_point: &str, append_to_path: bool) {}
                                ^
```

what it should do
```
before M-b

pub fn mount(&self, path: &str, mount_point: &str, append_to_path: bool) {}
                                           ^

after M-b:

pub fn mount(&self, path: &str, mount_point: &str, append_to_path: bool) {}
                                     ^
```

roysc:
By passing 'symbols as 2nd argument to regexp-opt, keywords will not be
erroneously matched inside symbols. This obviates changing the syntax of
`_' to "w".

Fixes https://github.com/rust-lang/rust-mode/issues/41
Fixes #11

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -128,7 +128,7 @@
           ((skip-dot-identifier
             (lambda ()
               (when (looking-back (concat "\\." rust-re-ident))
-                (backward-word 1)
+                (forward-thing 'symbol -1)
                 (backward-char)
                 (- (current-column) rust-indent-offset)))))
         (cond
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -32,11 +32,6 @@
     (modify-syntax-entry ?\" "\"" table)
     (modify-syntax-entry ?\\ "\\" table)
 
-    ;; mark _ as a word constituent so that identifiers
-    ;; such as xyz_type don't cause type to be highlighted
-    ;; as a keyword
-    (modify-syntax-entry ?_ "w" table)
-
     ;; Comments
     (modify-syntax-entry ?/  ". 124b" table)
     (modify-syntax-entry ?*  ". 23"   table)
@@ -274,10 +269,10 @@
   (append
    `(
      ;; Keywords proper
-     (,(regexp-opt rust-mode-keywords 'words) . font-lock-keyword-face)
+     (,(regexp-opt rust-mode-keywords 'symbols) . font-lock-keyword-face)
 
      ;; Special types
-     (,(regexp-opt rust-special-types 'words) . font-lock-type-face)
+     (,(regexp-opt rust-special-types 'symbols) . font-lock-type-face)
 
      ;; Attributes like `#[bar(baz)]` or `#![bar(baz)]` or `#[bar = "baz"]`
      (,(rust-re-grab (concat "#\\!?\\[" rust-re-ident "[^]]*\\]"))


MicahChalmer:
Hmm.  It's not obvious to me why this change would break the method chain indentation feature in all three emacs versions, and the raw string handling on emacs 23 only.  On the surface it would seem to be unrelated to both of these things.  But that's why the tests are there...

In addition to fixing that, it'd be helpful to add tests to this PR to check the forward-word and backward-word behavior along with fixing it.

huonw:
The chain indentation is probably broken because `rust-align-to-method-chain` calls `backward-word`: https://github.com/rust-lang/rust-mode/blob/b15b0eb15da170610784fde7eaa288d39e63f723/rust-mode.el#L136

It's not obvious why the raw string handling is failing, but several tests seem to be failing because `fn main` is being treated as `fn in`.

pnkfelix:
Hmm, I suspect we may want to introduce a differentiation between `ident` and `word` (if there is not already one present), and then have the method-chaining code jdo `backward-ident` instead of `backward-word`.

roysc:
@pnkfelix I think `(forward-symbol -1)` should accomplish that. `word` vs. `symbol` generally is that distinction in Emacs (which is the point of this PR)

roysc:
Using `(forward-thing 'symbol -1)` seemed to fix some of the failed tests. However, I have no idea how to interpret the failed raw string tests.

huonw:
Those emacs23 failures are strange!

r? @pnkfelix (transferring reviewership, don't have the bandwidth right now.)

gsingh93:
Any update on this? Without this patch, some of the features of `syntax-subword-mode` are broken, as they require `_` to be a symbol character, not a word character.

rust-mode.el
```diff
@@ -323,10 +318,10 @@
   (append
    `(
      ;; Keywords proper
-     (,(regexp-opt rust-mode-keywords 'words) . font-lock-keyword-face)
+     (,(regexp-opt rust-mode-keywords 'symbols) . font-lock-keyword-face)
 
      ;; Special types
-     (,(regexp-opt rust-special-types 'words) . font-lock-type-face)
+     (,(regexp-opt rust-special-types 'symbols) . font-lock-type-face)
```
MicahChalmer: I figured out why the test fails in emacs 23: emacs 23 did not recognize passing `'symbols` to `regexp-opt`, only `'words`.  In emacs 23 this is treated as if you passed `t`, which only groups without surrounding it with `\_<` / `\_>`.  So it starts highlighting keywords in the middle of symbols--for instance it will highlight the "in" as a keyword in `fn main()`.

Replace the above lines with this or something like it, and it should work on all tested versions of emacs:

```
     ;; Keywords proper
     (,(concat "\\_<" (regexp-opt rust-mode-keywords t) "\\_>") . font-lock-keyword-face)

     ;; Special types
     (,(concat "\\_<" (regexp-opt rust-special-types t) "\\_>") . font-lock-type-face)
```

pnkfelix:
@roysc should I take over this PR?  It looks like @MicahChalmer has dissected the emacs23 issues with it as currently written...

roysc:
@pnkfelix Sure, no problem

pnkfelix:
closing in favor of #56.

PR #43, opened 2015-02-23
Indent inside strings after line-ending backslash

### LINKED ISSUE #9: emacs mode indendation is wrong for string constants with backslashes
<a href="https://github.com/nikomatsakis"></img></a> **Issue by [nikomatsakis](https://github.com/nikomatsakis)**
_Wednesday Apr 16, 2014 at 22:20 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/13568_

----

the following code indents oddly:

```
pub fn foo() {
    format!("abc \
def")
}
```

I expect:

```
pub fn foo() {
    format!("abc \
             def")
}
```

MicahChalmer:
Fix #9.

Note there are a few different ways it tries to determine how to indent a line within a string.  Here are some examples from the tests:

```rust
pub fn aligned_with_str_begin() {
    format!("abc \
             def");
}


pub fn indented_to_next_level() {
    format!("\
        abc \
        def");
}

pub fn arbitrary_indents_with_and_without_backslashes() {
    
    println!("
Here is the beginning of the string
            and here is a line that is arbitrarily indented \
            *and a continuation of that indented line
     and another arbitrary indentation
  still another
        yet another \
        *with a line continuing it
And another line not indented
");

}
```

In the first function, the continuation aligns with the start of the string.  In the second, it pushes out to the next indentation level, since there is nothing in the string on the previous line.  (This is intended to be similar to how it indents lists inside parentheses--aligns to the open paren if there is anything following it on the same line, indent to next level otherwise.)

If the line ending with a backslash is not the first line in the string, the indent level of that line is used to continue it.  In the third function, the indentation will change the whitespace before the lines beginning in `*` in order to align them with their previous lines.  It will leave all the other lines of that string entirely untouched.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -117,7 +117,7 @@
     ;; be undone via tab.
     
     (when (looking-at (concat "\s*\." rust-re-ident))
-      (previous-logical-line)
+      (forward-line -1)
       (end-of-line)
 
       (let
@@ -176,7 +176,7 @@
                    ((string-begin-pos (nth 8 (syntax-ppss)))
                     (end-of-prev-line-pos (when (> (line-number-at-pos) 1)
                                             (save-excursion
-                                              (previous-line)
+                                              (forward-line -1)
                                               (end-of-line)
                                               (point)))))
                  (when
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -164,10 +164,58 @@
                       (when rust-indent-method-chain
                         (rust-align-to-method-chain))
                       (save-excursion
+                        (rust-rewind-irrelevant)
                         (backward-up-list)
                         (rust-rewind-to-beginning-of-current-level-expr)
                         (+ (current-column) rust-indent-offset))))))
              (cond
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -164,10 +164,58 @@
...
                         (rust-align-to-method-chain))
                       (save-excursion
+                        (rust-rewind-irrelevant)
                         (backward-up-list)
                         (rust-rewind-to-beginning-of-current-level-expr)
                         (+ (current-column) rust-indent-offset))))))
              (cond
+              ;; Indent inside a non-raw string only if the the previous line
+              ;; ends with a backslash that is is inside the same string
+              ((nth 3 (syntax-ppss))
```
nikomatsakis: how "well known and unchangable" are these indices? it feels like it'd be more obvious what's happening here if we at least used symbolic names.

nikomatsakis: oh, I see, it's returning into some crazy tuple...ok.

PR #49, opened 2015-03-29
Add rust-playpen-buffer and rust-playpen-region for sharing code.

huonw:
This will create (and shorten) a URL with the code on
play.rust-lang.org.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -67,6 +67,15 @@
   :type 'boolean
   :group 'rust-mode)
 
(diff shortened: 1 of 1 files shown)


huonw:
I don't know if `message` is the best way to inform the user of the shortened URL.

huonw:
I also don't like the `last-line` hack, which is working around the fact that the buffer created by `url-retrieve` includes all the HTTP headers.

PR #52, opened 2015-04-14
Fix "-> Type" on a single line indentation

### LINKED ISSUE #8: emacs mode indentation is wrong when function return -> is on its own line
<a href="https://github.com/nikomatsakis"></img></a> **Issue by [nikomatsakis](https://github.com/nikomatsakis)**
_Wednesday Apr 16, 2014 at 19:39 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/13566_

----

The following code indents as shown here, which is weird:

```
pub fn from_pointer_kind(base_mutbl: MutabilityCategory,
                         ptr: PointerKind)
                         -> MutabilityCategory {
                             match ptr {
                             }
                         }
```

I expect it to be:

```
pub fn from_pointer_kind(base_mutbl: MutabilityCategory,
                         ptr: PointerKind)
                         -> MutabilityCategory {
    match ptr {
    }
}
```

GBGamer:
Plus a few minor spelling changes in comments

Fixes #8

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -102,6 +102,9 @@
 (defun rust-rewind-to-beginning-of-current-level-expr ()
   (let ((current-level (rust-paren-level)))
     (back-to-indentation)
+    (when (looking-at "->")
+      (rust-rewind-irrelevant)
+      (back-to-indentation))
     (while (> (rust-paren-level) current-level)
       (backward-up-list)
       (back-to-indentation))))
@@ -179,7 +182,7 @@
                         (+ (current-column) rust-indent-offset))))))
              (cond
               ;; Indent inside a non-raw string only if the the previous line
-              ;; ends with a backslash that is is inside the same string
+              ;; ends with a backslash that is inside the same string
               ((nth 3 (syntax-ppss))
                (let*
                    ((string-begin-pos (nth 8 (syntax-ppss)))
@@ -232,7 +235,7 @@
                  (or (rust-align-to-expr-after-brace)
                      (+ baseline rust-indent-offset))))
 
-              ;; A closing brace is 1 level unindended
+              ;; A closing brace is 1 level unindented
               ((looking-at "}") (- baseline rust-indent-offset))
 
               ;; Doc comments in /** style with leading * indent to line up the *s


GBGamer:
Can I just say thank you to the people who wrote this? Incredibly well commented, easy to edit, awesomeness.

quantheory:
Ah! This is definitely an improvement. I find that there are a *lot* of other cases where indentation is at least somewhat strange in `rust-mode`, which is perhaps an argument for switching to a more "tried and true" way of dealing with indentation, e.g. based on SMIE. (To see what I mean, find a random `if` statement with a moderately complex condition, where the `{` is after the condition rather than on a line by itself, and insert a newline between any random pair of tokens, then reindent the block. This will almost certainly mess up the indentation of the entire `if` block. Several other types of expression are similarly fragile to users adding newlines in "unexpected" places.)

Nevertheless, this PR works around one of the most common bad cases, so it is still quite helpful.

GBGamer:
Yes, we can work on the rest of the indentation issues later, this is just an issue that I had that was an easy fix.

GBGamer:
@quantheory I'm setting up a new repo in order to work on an SMIE grammar for Rust. Would you like to join me? (I've never done any SMIE work before (and in fact, am an Emacs newbie), so I'd like someone who actually knows what they're doing.

quantheory:
@GBGamer I would, but I'm afraid I am really the wrong person to ask. I actually only have a dimly-remembered mental picture of how SMIE works from several years back, and I'm pretty overextended in any case. I really only check in here because I like to ~~whine~~ hope that some of the quirks in the current `rust-mode` will get ironed out soon.

GBGamer:
Pinging @nikomatsakis

pnkfelix:
Okay, normally I'd say "please add a test case."

But I want this badly enough that I'm probably going to just merge it.  (I just want to quickly double-check myself that it behaves as advertised -- if so, then I'll merge it and then work on a test case myself.)

pnkfelix:
Yes, awesome, it fixes both of my most important cases:

```rust
    fn foo(a: i32)
           -> i32 {
               // whoops
           }
                                                                                
    fn foo<T>(a: i32)
              -> i32 where T:Foo {
                  // bad too
              }
```

GBGamer:
@pnkfelix Thanks for the regression test, I didn't even think about that :stuck_out_tongue_closed_eyes:

PR #56, opened 2015-04-30
Fix word and symbol syntax distinction, take 2

### LINKED ISSUE #11: Emacs rust-mode: backward-word and backward-kill broken

### LINKED ISSUE #41: {forward,backward}-word behavior is unexpected

pnkfelix:
Adapted from @roysc's PR #42.

Do not match keywords inside symbols when building the `rust-mode-font-lock-keywords`. This obviates changing the syntax of `_' to "w", fixing other problems.

Fix #41
Fix #11

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -140,7 +140,7 @@
           ((skip-dot-identifier
             (lambda ()
               (when (looking-back (concat "\\." rust-re-ident))
-                (backward-word 1)
+                (forward-thing 'symbol -1)
                 (backward-char)
                 (- (current-column) rust-indent-offset)))))
         (cond
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -326,14 +326,19 @@
 (defun rust-re-item-def (itype)
   (concat (rust-re-word itype) "[[:space:]]+" (rust-re-grab rust-re-ident)))
 
+;; (See PR #42 -- this is just like `(regexp-opt words 'symbols)` from
+;; newer Emacs versions, but will work on Emacs 23.)
+(defun regexp-opt-symbols (words)
+  (concat "\\_<" (regexp-opt words t) "\\_>"))
+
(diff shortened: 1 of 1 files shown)


pnkfelix:
(treating self as reviewer; all I did was adopt the patches of other people, as documented on #56.)

PR #64, opened 2015-05-18
Fontifies documentation comments

### LINKED ISSUE #61: Documentation String Highlight
It would be nice if documentation strings were highlight a different colour to the usual comment strings.  Emacs' font-lock seems to support documentation strings through setting `font-lock-doc-face`.  

The vim analogue of rust-mode does this and highlights documentation in a very-visible colour (red in solarized dark), this makes reading the documentation really nice within the code.

davazp:
I have implemented fontification of both line (/// and //!) and block (/** and /*!) comments. It uses the `font-lock-doc-face`, which does not look very pretty in some themes. We could change this if necessary. It includes unit tests and it fixes #61.

See a couple of screenshots:

example

test

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -10,8 +10,10 @@
 
 ;;; Code:
 
-(eval-when-compile (require 'misc)
-                   (require 'rx))
+(eval-when-compile
+  (require 'misc)
+  (require 'rx)
+  (require 'cl))
 
 ;; for GNU Emacs < 24.3
 (eval-when-compile
@@ -386,12 +388,12 @@
   ;; whole comment will be.
   (let ((start font-lock-beg)
         (end font-lock-end))
-    (cl-loop for level = (syntax-ppss start)
-             while (and (integerp level) (plusp start))
-             do (decf start))
-    (cl-loop for level = (syntax-ppss end)
-             while (and (integerp level) (plusp end))
-             do (incf end))
+    (loop for level = (syntax-ppss start)
+	  while (and (integerp level) (plusp start))
+	  do (decf start))
+    (loop for level = (syntax-ppss end)
+	  while (and (integerp level) (plusp end))
+	  do (incf end))
     (unless (and (eql start font-lock-beg) (eql end font-lock-end))
       (setq font-lock-beg start)
       (setq font-lock-end end)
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -374,6 +379,51 @@
...
+             do (incf end))
+    (unless (and (eql start font-lock-beg) (eql end font-lock-end))
+      (setq font-lock-beg start)
+      (setq font-lock-end end)
+      t)))
+  
+
+(defun rust-look-for-line-doc-comment (bound)
+  ;; Find documentation comments starging with //! or ///. It will
+  ;; only match it if it is inside a non-nestable comment.
```
pnkfelix: typo: "starging" should be "starting"

rust-mode.el
```diff
@@ -374,6 +379,51 @@
...
+  (when (search-forward-regexp "//[!/].*" bound t)
+    (eq (nth 4 (syntax-ppss)) t)))
+
+
+(defun rust-look-for-block-doc-comment (bound)
+  ;; Find documentation comments starging with /** or /*!. The
```
pnkfelix: again, "starging"

davazp:
I use Emacs 24.4.1.  I haven't test it at all with other versions.

pnkfelix:
@davazp  I assume your response means that you noticed that travis CI says that emacs23 build is failing with this change?

davazp:
Actually I didn't notice, but I read about emacs23 in the README.md and I supposed things could go wrong.  It was easy to fix. I have switched `cl-loop` to `loop` (after requiring `cl` at compile time) and now it builds.

Wilfred:
I'd love to see this merged -- is it missing anything? Looks good to me, other than one minor comment.

rust-mode.el
```diff
@@ -374,6 +381,51 @@
              ("fn" . font-lock-function-name-face)
              ("static" . font-lock-constant-face)))))
 
+
+(defun rust-extend-region-nested-comment ()
+  ;; Extend the region given by `font-lock-beg' and `font-lock-end' to
```
Wilfred: I think these should be docstrings, what do you think?

tomjakubowski: I agree they should be docstrings.

I opened an issue to run checkdoc as part of the test suite, I think it's a must-have for any elisp package. #70

davazp:
I close this pull request as #71 is a better solution.

PR #65, opened 2015-05-22
handle case where blink-matching-paren-distance is nil

tromey:
If blink-matching-paren-distance is nil -- a valid value -- then rust-find-matching-angle-bracket will error.


nikomatsakis:
@tromey thanks!

PR #67, opened 2015-05-22
fix quoting in two doc strings

tromey:
This fixes the quotes in a couple of docstrings.


steveklabnik:
So, using a backtick and a `'` is a pretty common quoting style. That doesn't mean that we shouldn't switch it, but it should also change the backticks to `'`, if we are.

steveklabnik:
oh, i see. i read the diff wrong. sigh.

huonw:
I'll be @bors. :smile:

PR #69, opened 2015-05-24
set :safe on two defcustoms

tromey:
This sets the :safe property on a couple of defcustoms.  This lets users set these variables in file-local variables or in .dir-locals.el.


Wilfred:
Looks good to me, FWIW.

MicahChalmer:
These make sense to me too.  Merging it, thanks!

PR #71, opened 2015-06-07
Distinguish face for doc-comments

birkenfeld:
Using the syntactic-face-function, we can assign the proper
`font-lock-doc-face' to doc comments ("///" and "//!").

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1085,6 +1085,70 @@ this_is_not_a_string();)"
     (should (equal nil (get-text-property 28 'face))) ;; Semicolon--should not be part of the string
     ))
 
+;;; Documentation comments
+
+(ert-deftest font-lock-doc-line-comment-parent ()
+  (rust-test-font-lock
+   "//! doc"
+   '("//! doc" font-lock-doc-face)))
+
+(ert-deftest font-lock-doc-line-comment-item ()
+  (rust-test-font-lock
+   "/// doc"
+   '("/// doc" font-lock-doc-face)))
+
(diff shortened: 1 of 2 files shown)


steveklabnik:
I just commented on your other PR as well, would you mind taking a screenshot or tossing the HTML up somewhere?

birkenfeld:
Actually this is similar to #64.

The implementation strategy for #64 is different (it uses entries in the font-lock keywords table instead of the syntactic face function). I am not enough of an emacs hacker to say which strategy is better, except that I adapted my patch from the Python and Haskell modes.

#64 has tests, I can roll them into my patch.

lunaryorn:
@birkenfeld This one is better, since it hooks into syntactic fontification to pick the appropriate face whereas the other one just re-fontifies doc comments.

Effectively, this strategy looks at comments only once, whereas the other one effectively fontifies each doc comment twice (first during syntactic fontification, and then again with font lock keywords).

davazp:
I also prefer this to my approach in #64. I did not know about `font-lock-syntactic-face-function`. Nice one!  I will close my pull request.

Just a note, we should not highlight comments like `/*** ... */`, it could highlight some header comments badly:
> and block comments beginning with **exactly one** repeated asterisk in the block-open sequence (/**)

pnkfelix:
This needs a rebase I think

birkenfeld:
Rebased, and added the fix for not highlighting `/***...*/` and `////...` as doc comments.

birkenfeld:
Thanks!

PR #72, opened 2015-06-07
Change font-lock face for module paths

birkenfeld:
The `font-lock-type-face' is usually quite bold, which is nice for
type constructors and type names in "struct X" or "impl X".  But
for module paths, it may be better to use a face that is usually
more muted.

Since `font-lock-builtin-face' is not used at the moment, I'm
proposing this.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1422,16 +1422,34 @@ this_is_not_a_string();)"
    "\"/*! doc */\""
    '("\"/*! doc */\"" font-lock-string-face)))
 
-(ert-deftest font-lock-module ()
+(ert-deftest font-lock-module-def ()
+  (rust-test-font-lock
+   "mod foo;"
+   '("mod" font-lock-keyword-face
+     "foo" font-lock-constant-face)))
+
+(ert-deftest font-lock-module-use ()
+  (rust-test-font-lock
+   "use foo;"
+   '("use" font-lock-keyword-face
+     "foo" font-lock-constant-face)))
+
+(ert-deftest font-lock-module-path ()
   (rust-test-font-lock
    "foo::bar"
-   '("foo" font-lock-type-face)))
+   '("foo" font-lock-constant-face)))
 
-(ert-deftest font-lock-submodule ()
+(ert-deftest font-lock-submodule-path ()
   (rust-test-font-lock
    "foo::bar::baz"
-   '("foo" font-lock-type-face
-     "bar" font-lock-type-face)))
+   '("foo" font-lock-constant-face
+     "bar" font-lock-constant-face)))
+
+(ert-deftest font-lock-type ()
+  (rust-test-font-lock
+   "foo::Bar::baz"
+   '("foo" font-lock-constant-face
+     "Bar" font-lock-type-face)))
 
 (ert-deftest font-lock-type-annotation ()
   "Ensure type annotations are not confused with modules."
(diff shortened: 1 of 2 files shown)


steveklabnik:
Hey @birkenfeld! Thanks for the PR.

Usually, for changes like this, we ask if you could toss up a rendered version of the docs, either a screenshot or the HTML so we could check it out. Any chance you could do that? If not, I'll pull it down and try myself.

birkenfeld:
@steveklabnik Sure, although how the screenshot looks is heavily dependent on the color theme in use.

birkenfeld:
Base version:
a
With patch:
b

birkenfeld:
Rebased...

MicahChalmer:
This doesn't actually distinguish module names from types--it just highlights everything but the last part of a chain joined by `::` differently.  Take `fern::OutputConfig` in the example above--`OutputConfig` is presumably a type, but it's highlighted as a module path in `fern::OutputConfig::stdout()`.  Same thing with `Server` in `server::Server::new`.

I wouldn't expect rust-mode to actually be able to tell the difference.  `Server` and `OutputConfig` _could_ be modules--they'd just get a warning for having the wrong case conventions.  But since rust-mode can't really make the distinction, does coloring all but the last identifier differently add useful information?

birkenfeld:
You're right, of course. One could make a heuristic distinction based on the upper/lowercasing of the path components.

But the main argument is that the important part is the last one, so it deserves to stand out more IMO (as I said, `font-lock-type-face` is usually quite eye-catching).

MicahChalmer:
I'm looking at `cc-mode`'s C++ highlighting for precedent.  (Given that Rust is targeted at C++'s use cases, it seems particularly relevant.)  The C++ mode that comes with `cc-mode` does something similar to this, but it uses `font-lock-constant-face` for namespaces rather than `font-lock-builtin-face`.  It also highlights namespace names with this face when they are declared as well.  It too can't distinguish types from namespaces inside qualified identifiers--C++ doesn't even have a strong capitalization convention around it like Rust does, so there wouldn't be any way to do so.  But for the same reason, it doesn't try to highlight types when they're used in general, only when they're declared.

Currently `rust-mode` always highlights modules as types.  It does so when they are declared and when they are used.  It also uses Rust's strongly-encouraged convention of types being capitalized, plus its own list of `rust-special-types`, to highlight identifiers that look like type names wherever they appear.  I think highlighting types in all places is a good thing, and would like it to continue even if we start distinguishing modules.  Highlighting module names differently than types would also be good, but I would like it if it did so on both declarations and uses.

In light of all this, this is what I think it should do:

  * Use `font-lock-constant-face` for module names in paths (in `time::now()`, `time` gets `font-lock-constant-face)
  * Highlight module names with `font-lock-constant-face` when declared as well (in `mod server`, `server` gets `font-lock-constant-face`)
  * Still highlight types (as determined via capitalization or membership in `rust-special-types`) using `font-lock-type-face` when they appear in paths (so in `fern::OutputConfig::stdout()`, `fern` gets `font-lock-constant-face` and `OutputConfig` gets `font-lock-type-face`)

birkenfeld:
Sounds good!

nikomatsakis:
So I just realized this PR has been hanging for...almost a year? I'm unclear on what the status is here :)

birkenfeld:
I've updated the PR to implement @MicahChalmer 's plan on top of the latest master.

MicahChalmer:
Cool...looks good to me with the update.

PR #77, opened 2015-06-17
fix a bug; then prevent reintroduction of similar bugs

tromey:
This series starts by fixing a bug I found.  An unbound variable was referenced.

Then it proceeds to make such bugs harder to introduce, by requiring a clean byte-compile; and then by using lexbind to get even better checking from the compiler.

### DIFF
--- a/run_rust_emacs_tests.sh
+++ b/run_rust_emacs_tests.sh
@@ -31,4 +31,13 @@ $( $EMACS -batch --eval "(require 'ert)" > /dev/null 2>&1 ) || {
     exit 3
 };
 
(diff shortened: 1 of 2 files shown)


tromey:
Note that this pull request may need a tweak depending on whether #78 goes in first.

tromey:
This one will require some changes given recent pulls.

MicahChalmer:
I definitely agree with the ideas here.  @tromey if you rebase/fix this, I'd like to merge it.

tromey:
@MicahChalmer - I've updated the branch.

PR #78, opened 2015-06-21
test $EMACS directly rather than using "which"

### LINKED ISSUE #33: test driver `which` test disallows useful values for `$EMACS`
Spawned off of PR #31:

@nikomatsakis said: 
> As a side note I was annoyed by the "which" test because I wanted to use EMACS="emacs23 --no-site-init" and it was not allowed. Since this runnable test seems to cover the which test maybe we can remove that.

tromey:
Fixes #33


huonw:
Thanks!

PR #79, opened 2015-06-25
Match angle brackets syntactically

### LINKED ISSUE #10: Adding a line break after a Trait in a template argument list breaks emacs-mode indentation
<a href="https://github.com/gsingh93"></img></a> **Issue by [gsingh93](https://github.com/gsingh93)**
_Friday Aug 22, 2014 at 01:25 GMT_

_For earlier discussion, see https://github.com/rust-lang/rust/issues/16665_

----

I think this is a somewhat serious bug with emacs-mode, since I can't use the tab key for indentation inside a function block in many cases.

My code ends up with a long list of traits with bounds in the template argument list, and according to the [style guide](https://aturon.github.io/style/whitespace.html#line-wrapping), I can wrap between traits. However, if I wrap between traits, identation is broken for the function. Consider the following example,
```
pub fn foo<T,
           V>() {
              // Indentation puts me here
}
```
As you can see, hitting tab inside the function idents too much. I should mention that I aligned the second template argument to match up with the first, as hitting tab there doesn't work either.

### LINKED ISSUE #50: Angle bracket matching doesn't handle `->`
If you type `Box<FnBox() -> i32>`, the trailing `>` doesn't match properly. This should be easy enough to fix!

MicahChalmer:
Remove the angle bracket blinking feature, and make it actually match `<` and `>` characters as parens in the syntax table instead.  This matching is suppressed as needed, so that less-than and greater-than operators, arrow operators, etc don't throw off the matches.

This has the following effects:

  * Angle brackets match like other braces/parens however they are configured, rather than blinking in a different way.
  * Movement commands like `forward-sexp` and `backward-up-list` acknowledge angle brackets.
  * `electric-pair-mode` works with angle brackets (also suppressed when needed.)
  * Fix #50 - angle brackets are matched in the right pairs; `->` and other non-angle-bracket `<` and `>` characters do not attempt to match.
  * Fix #10 - multi-line lists inside angle brackets are indented like other lists.

Macros have arbitrary syntax and therefore it's not possible to know whether a `<` or `>` character is an angle bracket or not.  So the matching is always suppressed inside macro definitions and uses.

There are a few disadvantages: besides the fact that the code is rather hairy, it also has to detect the non-angle-brackets perfectly, or else it throws off the indentation for the whole rest of the file.  For that reason there is still a custom variable available to disable it (though with a different name than the previous one.)

Having said that, I don't know of any cases where it gets it wrong.  I tested it with [this program](https://gist.github.com/MicahChalmer/1ac9f8de543655e577a1), which compares what emacs thinks are the non-bracket `<`/`>` characters with what the rust compiler thinks.  It passed for all of the rust compiler code base, as well as all the rust code in servo and as many of servo's dependencies as still compile with the latest nightly.

Also included are some bug fixes for some edge case issues in the handling of raw strings and character literals.  I found these while running the test script.  Even if the angle bracket matching itself is deemed to be too risky or hairy to be merged (and I would understand it if that turns out to be the case), you may want to merge the other commits for the bug fixes.  I've put them first to make this easy.

If this PR is accepted, close #66 (because the variable that it renames has been removed entirely) and close #74 (because post-self-insert-hook is no longer used at all.)  Both of these pertain to the old angle bracket blinking code that this PR removes entirely.

And if the PR is not accepted - hey, I still had fun writing it.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -710,6 +710,7 @@ INIT-POS, FINAL-POS are position symbols found in `rust-test-positions-alist'."
   (with-temp-buffer
     (rust-mode)
     (insert source-code)
+    (font-lock-fontify-buffer)
     (goto-char (rust-get-buffer-pos init-pos))
     (apply manip-func args)
     (should (equal (point) (rust-get-buffer-pos final-pos)))))
@@ -723,6 +724,7 @@ All positions are position symbols found in `rust-test-positions-alist'."
   (with-temp-buffer
     (rust-mode)
     (insert source-code)
+    (font-lock-fontify-buffer)
     (goto-char (rust-get-buffer-pos init-pos))
     (apply manip-func args)
     (should (equal (list (region-beginning) (region-end))
@@ -1303,7 +1305,7 @@ fn indented_already() {
     \n    // The previous line already has its spaces
 }
 ")
-
+    (font-lock-fontify-buffer)
     (goto-line 11)
     (move-to-column 0)
     (indent-for-tab-command)
@@ -1483,6 +1485,14 @@ la la\");
    ;; Needs to leave 1 space before "world"
    "\"hello \\\n world\""))
 
+(ert-deftest indent-multi-line-type-param-list ()
+  (test-indent
+   "
+pub fn foo<T,
+           V>() {
+    hello();
+}"))
+
 (defun rust-test-matching-parens (content pairs &optional nonparen-positions)
   "Assert that in rust-mode, given a buffer with the given `content',
   emacs's paren matching will find all of the pairs of positions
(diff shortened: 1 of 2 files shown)


pnkfelix:
I think I do want to accept this.  At the very least, I've noticed that the current bracket-blinking code totally messes up multiple-cursors mode.

cc @nikomatsakis any thoughts here?

MicahChalmer:
Thanks @pnkfelix.

I just rebased after the latest other commits to master.  (Nothing else was changed.)

nikomatsakis:
I have no objections -- it sounds strictly better than the hack I did. I don't quite understand how it works (but I haven't tried reading it).

PR #82, opened 2015-07-09
Correctly indent where clauses

mrBliss:
When you put `where` on a new line along with the opening brace of the function, the indentation of the body was off. To circumvent this wrong indentation, you'd put the opening brace on a new line, but that goes against the style guide. The first commit corrects the wrongful indentation.

This doesn't seem to be mentioned in the Rust style guide, but multi-line `where`-clauses are often aligned on the type parameters. The second commit takes care of this.

See the commit messages and unit tests for examples.

Thanks for having a good test suite, this helped me tremendously!

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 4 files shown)


mrBliss:
Is there any reason this hasn't been merged after three weeks?

MicahChalmer:
With all the discussion going on in rust-lang/rfcs#1190 it seems unclear to me how rust-mode should really format these.  One problem (discussed in the link by others) is that if you indent the "where" as in the first example, then the "where" ends up indented the same amount as the body, losing the visual distinction between the body and the signature.

Perhaps once there is a chosen rustfmt default style, rust-mode's indentation can then be made to implement it.

On the other hand, what it does now is completely incoherent, so there's a case for merging something like this even before the question is fully resolved.

PR #84, opened 2015-07-15
Recognize compiler notes as "info" messages

birkenfeld:
Otherwise they will be matched as errors by less-specific regexps
built into compilation mode.

Also uses the previously unused let-definition of the error/warning string.


MicahChalmer:
This now needs a rebase.  Ideally, there should also be tests added for this.

PR #85, opened 2015-07-19
Cache (regexp-opt rust-special-types 'symbols) in rust-is-lt-char-operator

talchas:
regexp-opt is generally intended to be cached, and the extra GC was
causing noticeable lag when typing sometimes.

I still get lag in some cases from rust-in-macro calling syntax-ppss and falling through to parse-partial-sexp all the time, but fixing that in a not-horrible way is less trivial.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -378,6 +378,7 @@
 ;; newer Emacs versions, but will work on Emacs 23.)
 (defun regexp-opt-symbols (words)
   (concat "\\_<" (regexp-opt words t) "\\_>"))
+(defconst rust-re-special-types (regexp-opt-symbols rust-special-types))
 
 (defvar rust-mode-font-lock-keywords
   (append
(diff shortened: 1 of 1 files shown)


pnkfelix:
@talchas should we be putting in similar `defconst` definitions for the other regexps that are currently generated inline via `regexp-opt` elsewhere in that code?  Or is it just `rust-special-types` that is getting used in a particularly hot loop here?

talchas:
All the other ones were already like that, so they're fine. (The only other serious perf issue I've run into is the syntax-ppss one, for which my current hack is to set syntax-ppss-stats so that it isn't insane (which is emacs' fault)).

pnkfelix:
@talchas oh I see, I was quickly skimming and saw the two calls to `regexp-opt-symbols` in the lines below your `defconst`, but those are only in a `defvar`, not in a `defun` the way that the code you replaced was.

Okay, looks good.

PR #88, opened 2015-07-24
Support nested block comments

kw-udon:
Support nested block comments like `/* /* ...  */ */`.


Wilfred:
I can confirm this works. @huonw is there anything else needed to merge this?

MicahChalmer:
It'd be good if the PR came with a test in rust-mode-tests.el.  But for something that passes the existing tests and changes so little I don't think that needs to be a blocker.

PR #89, opened 2015-07-31
Fix slow angle bracket matching

### LINKED ISSUE #86: Large numbers of <>'s in a file cause massive slowdown
E.g. it takes several seconds for a keypress to appear at the bottom of a file with 400 copies of `fn foo<T>() {}`. Editing the same file after `s/<T>/_T_/` (to preserve the size) is fine.

MicahChalmer:
Fix #86.

The problem was caused by the logic that made it refontify the whole buffer when "font-lock-syntactically-fontified" was set to a position before the start of the region to be potentially fontified. Unfortunately that variable is not reliably set when fontifying a large buffer.  Fortunately, the new logic is much simpler, and font-lock already takes care of ensuring that everything before font-lock-beg was syntactically fontified.

The other problem was calling syntax-ppss on positions known not to be fontified yet.

This fixes both of these issues, and the angle bracket matching now works on larger buffers without pausing on every keystroke.

Also included is a patch that makes it recognize incomplete raw strings.  Without this, changing a character can affect syntax before it, and font-lock can be glitchy if that happens.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1110,6 +1110,41 @@ this_is_not_a_string();)"
     (should (equal nil (get-text-property 28 'face))) ;; Semicolon--should not be part of the string
     ))
 
+(ert-deftest font-lock-runaway-raw-string ()
+  (rust-test-font-lock
+   "const Z = r#\"my raw string\";\n// oops this is still in the string"
+   '("const" font-lock-keyword-face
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Looked pretty reasonable to me. Thanks @MicahChalmer for looking into this!

ps, I'm always happy when I hit tab and things line up with the opening `<` :)

PR #105, opened 2015-10-23
Remove unused require of misc.el

### LINKED ISSUE #95: misc not provided
Error 'misc' is not provided, error thrown when opening *.rs files. I am assuming this is referring to line 13 in rust-mode.el. Any help is welcome, Thanks

MicahChalmer:
Fix #95. I don't know what exactly it is that causes the issue, since as far as I know `misc.el` is provided with the version of emacs that people are seeing the issue #95 in.  But requiring it in `rust-mode.el` is no longer needed, and it's causing some people problems, so let's get rid of it.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -10,8 +10,7 @@
 
 ;;; Code:
 
-(eval-when-compile (require 'misc)
-                   (require 'rx)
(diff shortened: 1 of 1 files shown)


nikomatsakis:
lgtm

PR #109, opened 2015-11-26
Highlight the unsafe keyword

mrBliss:
As suggested by a recent [Reddit thread](https://www.reddit.com/r/rust/comments/3u9wxu/tip_use_a_special_color_for_the_unsafe_keyword/).

I've added a custom face (`rust-unsafe-face`) that inherits from `font-lock-warning-face` for easy customisation.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -173,6 +173,11 @@
   :safe #'booleanp
   :group 'rust-mode)
 
+(defface rust-unsafe-face
+  '((t :inherit font-lock-warning-face))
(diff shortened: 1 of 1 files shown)


nikomatsakis:
cc @MicahChalmer I don't know enough about "faces" and the like to know if this is best practice, but seems overall fine to me.

MicahChalmer:
The way the face is set up here looks right to me. It's based on an existing one so that normal color schemes work, but you can customize this one in particular if you want to. Let's merge it.

PR #111, opened 2015-12-02
Correctly indent where clauses

mrBliss:
Revisit #82. Now rustfmt has a default style for `where` clauses, it
makes sense to let rust-mode support it by default. How rust-mode
currently indents `where` clauses is totally broken anyway. The line
starting with `where` itself, the following lines that are part of the
`where` clause, and the body were all indented incorrectly. This commit
fixes this.

**Note that this commit does not prescribe a certain indentation style
for where clauses, but supports the most common ones, including
rustfmt's default.**

By choosing the location of (1) `where` and (2) the type parameter
bounds, the user can follow three major indentation styles. There is no
need for a configuration option to choose the style, as the locations of
1 and 2 dictate the style to follow. So all three styles can be used
together in the same file.

For each major style, the opening brace can be on the last line of the
type parameter bounds or on a new line (a or b in the examples). All 6
styles are supported without having to change any configuration option.
See the examples below and the new test cases.

There is one more style that is unfortunately incompatible with the
default `where` indentation style of rustfmt: when the user does not
want `where` to be indented, but aligned with the `fn` or `trait` on the
line before (see examples `foo4a` and `foo4b` below). To enable this
style, the user has to set the new option `rust-indent-where-clause` to
nil (it defaults to t).

Putting `where` and the type parameter bounds on the same line as the
function header is still supported of course (see examples `fooa` and
`foob` below). As there is no indentation, this commit does not
influence this.

Note that rust-mode's indentation has a different goal than rustfmt.
rustfmt reflows code, adds or removes line breaks, reindents, etc.
rust-mode's indentation only indents lines. The goal is not to imitate
rustfmt, but after running rustfmt on a file, reindenting it with
rust-mode should be idempotent. This way, users do not have to deal with
the eternal battle of differing indentation styles of rustfmt and
rust-mode.

The supported styles:

```rust
fn foo1a(a: A, b: B) -> C
    where A: Clone + Default,
          B: Eq,
          C: PartialEq {
    let body;
}

fn foo1b(a: A, b: B) -> C
    where A: Clone + Default,
          B: Eq,
          C: PartialEq
{
    let body;
}
```

```rust
fn foo2a(a: A, b: B) -> C where A: Clone + Default,
                                B: Eq,
                                C: PartialEq {
    let body;
}

fn foo2b(a: A, b: B) -> C where A: Clone + Default,
                                B: Eq,
                                C: PartialEq
{
    let body;
}
```

```rust
fn foo3a(a: A, b: B) -> C where
    A: Clone + Default,
    B: Eq,
    C: PartialEq {
    let body;
}

fn foo3b(a: A, b: B) -> C where
    A: Clone + Default,
    B: Eq,
    C: PartialEq
{
    let body;
}
```

If the user wants `foo4` instead of `foo1`, `rust-indent-where-clause`
must be set to nil. `foo2` and `foo3` are not influenced by this.

```rust
fn foo4a(a: A, b: B) -> C
where A: Clone + Default,
      B: Eq,
      C: PartialEq {
    let body;
}

fn foo4a(a: A, b: B) -> C
where A: Clone + Default,
      B: Eq,
      C: PartialEq
{
    let body;
}
```

Unchanged:

```rust
fn fooa(a: A, b: B) -> C where A: Clone + Default, ... {
    let body;
}

fn foob(a: A, b: B) -> C where A: Clone + Default, ...
{
    let body;
}
```

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -443,6 +443,166 @@ fn foo4(a:int,
 }
 "))
 
+(ert-deftest indent-body-after-where ()
+  (test-indent
+   "
+fn foo1(a: A, b: B) -> A
+    where A: Clone + Default, B: Eq {
+    let body;
+    Foo {
+        bar: 3
+    }
+}
+
+fn foo2(a: A, b: B) -> A
+    where A: Clone + Default, B: Eq
+{
+    let body;
+    Foo {
+        bar: 3
+    }
+}
+"))
+
+(ert-deftest indent-align-where-clauses-style1a ()
+  (test-indent
+   "
+fn foo1a(a: A, b: B, c: C) -> D
+    where A: Clone + Default,
+          B: Eq,
+          C: PartialEq,
+          D: PartialEq {
+    let body;
+    Foo {
+        bar: 3
+    }
+}
+"))
+
+(ert-deftest indent-align-where-clauses-style1b ()
+  (test-indent
+   "
+fn foo1b(a: A, b: B, c: C) -> D
+    where A: Clone + Default,
+          B: Eq,
+          C: PartialEq,
+          D: PartialEq
+{
+    let body;
+    Foo {
+        bar: 3
+    }
+}
+"))
+
+(ert-deftest indent-align-where-clauses-style2a ()
+  (test-indent
+   "
+fn foo2a(a: A, b: B, c: C) -> D where A: Clone + Default,
+                                      B: Eq,
+                                      C: PartialEq,
+                                      D: PartialEq {
+    let body;
+    Foo {
+        bar: 3
+    }
+}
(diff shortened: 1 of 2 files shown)


pnkfelix:
> The goal is not to imitate rustfmt, but after running rustfmt on a file, reindenting it with rust-mode should be idempotent.

Hmm, this sounds like a good overall property for rust-mode. It might be interesting to actually make its test-suite validate that this property holds for a set of inputs.

(Of course that has nothing to do with this patch has provided, its just an aside.)

nikomatsakis:
Nice!

pnkfelix:
> Nice!

(yes, by the way, I *love* our community!!!)

PR #113, opened 2015-12-15
Re-fontify buffer after it is reverted

### LINKED ISSUE #104: broken indentation in large files over time
I'm sorry I can't give a more specific error report, but I find pretty consistently that editing the files in `librustc_mir`, such as:

    build/mod.rs
    tcx/expr.rs

results in broken indentation. The failures modes seem to be one of two things:

1. pressing tab moves the text way over somewhere to the right for no apparent reason
2. I get errors about mismatched parentheses

Annoyingly, when I copy and paste the text into a new buffer to try and narrow it down, the problems often go away -- even if I copy everything. Closing and re-opening the buffer also makes things work again, as does disabling angle-bracket matching. This seems to suggest some sort of state that is being corrupted. @MicahChalmer you may have a good guess as to what this state is.

MicahChalmer:
Fix #104.  This makes it re-fontify the buffer after a revert, ensuring that it applies the right syntax properties to all the `<` and `>` characters.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -2427,6 +2427,29 @@ Fontification needs to include this whole string or none of it.
       (should (<= font-lock-beg 1))
       (should (>= font-lock-end 12)))))
 
+(ert-deftest redo-syntax-after-change-far-from-point ()  
+  (let*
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Nice!

PR #115, opened 2015-12-30
Fix type annotations incorrectly highlighted as modules.

Wilfred:
Previously, we were always treating `::` as a module, but Rust
allows type annotations using `::` e.g.

    parse::<i32>();


Wilfred:
My solution doesn't quite work with `foo::bar::baz`, will reopen once I've fixed it.

PR #116, opened 2016-01-01
Fix type annotations incorrectly highlighted as modules.

Wilfred:
Previously, we were always treating `::` as a module, but Rust
allows type annotations using `::` e.g.

    parse::<i32>();

This also changes module highlighting so that only the module name is
highlighted, excluding the `::`. This makes rust-mode consistent with
other Emacs modes, such as c++-mode and ruby-mode.

Before:

rust_before

After:

rust_after

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1403,6 +1403,27 @@ this_is_not_a_string();)"
    "\"/*! doc */\""
    '("\"/*! doc */\"" font-lock-string-face)))
 
+(ert-deftest font-lock-module ()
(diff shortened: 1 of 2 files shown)


Wilfred:
@nikomatsakis I've pushed a fix for highlighting of `foo:: <bar>` and added a matching test.

I've also refactored the while loop because I thought it was a little hard to read.

Let me know what you think or if you'd like any other changes. :)

nikomatsakis:
@Wilfred seems good to me! thanks.

PR #121, opened 2016-02-14
Integrate rustfmt support

### LINKED ISSUE #120: Integrate support for rustfmt
I think it would be great to have support for rustfmt integrated into the default rust-mode package. I think it should be bound to some hot-key and (not by default) configurable to run on every save. However, it looks like @fbergroth has already done [all the hard work](https://github.com/fbergroth/emacs-rustfmt/). Perhaps we can merge these two projects?

I wasn't sure where to open this issue, but I decided to do it here so that people could weigh in on whether they would like this support at all. But please @fbergroth I'd like to know your take as well -- would you prefer to keep emacs-rustfmt distinct?

fbergroth:
Merges the [rustfmt package](https://github.com/fbergroth/rust-mode).

Fixes #120.

Not sure if `C-c C-f` is the idiomatic binding, I use evil mode.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -181,6 +181,15 @@ function or trait.  When nil, where will be aligned with fn or trait."
   :safe #'booleanp
   :group 'rust-mode)
 
+(defcustom rust-format-on-save nil
+  "Format future rust buffers before saving using rustfmt."
+  :type 'boolean
+  :safe #'booleanp)
+
+(defcustom rust-rustfmt-bin "rustfmt"
+  "Path to rustfmt executable."
+  :type 'string)
+
 (defface rust-unsafe-face
   '((t :inherit font-lock-warning-face))
   "Face for the `unsafe' keyword."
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -1211,13 +1220,57 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
+    (rust--format-call (current-buffer))
+    (goto-char cur-point)
+    (set-window-start (selected-window) cur-win-start))
+  (message "Formatted buffer with rustfmt."))
+
+(defun rust-enable-format-on-save ()
+  "Enable formatting using rustfmt when saving buffer."
+  (interactive)
+  (add-hook 'before-save-hook #'rust-format-buffer nil t)
+  (message "Format on save enabled."))
```
mrBliss: Is this message really necessary? Idem for `rust-disable-format-on-save`.
I would find this rather annoying.

fbergroth: I agree. I could wrap it with `called-interactively-p`, so it wouldn't trigger on every buffer when `rust-format-on-save` is `t`. But probably the sane thing is just to remove it.

`"Formatted buffer with rustfmt"` will be displayed most often, you wouldn't find that annoying?

mrBliss: Wrapping it in `called-interactively-p` would be fine by me, removing it too.

It don't find `"Formatted buffer with rustfmt"` annoying, because it lets the user know the formatting actually took place. Users might get confused when formatting an already formatted file and not seeing anything happen (except for the modification indicator being set).

rust-mode.el
```diff
@@ -1211,13 +1220,57 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
+  (interactive)
+  (remove-hook 'before-save-hook #'rust-format-buffer t)
+  (message "Format on save disabled."))
+
 ;; For compatibility with Emacs < 24, derive conditionally
 (defalias 'rust-parent-mode
   (if (fboundp 'prog-mode) 'prog-mode 'fundamental-mode))
 
+(defvar rust-mode-map
+  (let ((map (make-keymap)))
```
mrBliss: Shouldn't this be `make-sparse-keymap`?

rust-mode.el
```diff
@@ -1211,13 +1220,57 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
+  (interactive)
+  (unless (executable-find rust-rustfmt-bin)
+    (error "Could not locate executable \"%s\"" rust-rustfmt-bin))
+
+  (let ((cur-point (point))
+        (cur-win-start (window-start)))
+    (rust--format-call (current-buffer))
+    (goto-char cur-point)
+    (set-window-start (selected-window) cur-win-start))
+  (message "Formatted buffer with rustfmt."))
```
mrBliss: This message is displayed even when formatting failed. It should be moved to the end of the `progn` in the first branch of the `if` in `rust--format-call` (after line 1231).

rust-mode.el
fbergroth: Indeed.

rust-mode.el
fbergroth: It shouldn't be displayed since failure is signalled with an `error`. Did you test it?

mrBliss: Ah, sorry, I didn't test it. You can ignore my comment.

fbergroth: NP :). I updated the commit with the other issues you addressed.

blaenk:
I noticed that clang-format primarily works on regions, and then `clang-format-buffer` is just `(clang-format-region (point-min) (point-max) style)`. That sounds like a nice way to implement it. I've seen a similar pattern with other things like gist.el.

While I currently can't think of a scenario where one would want to format just a specific region, the idea that that would be possible is very nice. Any chance we can do this? If not now, then after this has been merged? It seems like most of the work is already there.

fbergroth:
@blaenk Currently we can't do it since rustfmt has no support for formatting a specific part of the source, but it would be easy to support it on the emacs side once it's there.

Another thing that should be fixed at some is fbergroth/emacs-rustfmt#6.

blaenk:
Ah okay. I figured that it would probably break depending on the way you specified the region, e.g. in the middle of a parameter list or something, but I also figured it would work if you gave it a reasonable region no? Like if you have a huge file and you only completely select a single function definition, it wouldn't work? Oh well!

And yeah saving the position would be cool too. That's an interesting solution that gofmt takes in applying the patch. Shame it's not supported when reading from stdin. Is this a limitation, or it just hasn't been implemented in rustfmt I wonder?

fbergroth:
I think the sane way is to always send the entire source and making rustfmt allow optional parameters for start and end of the region to format. I doubt we'd get a satisfying solution by trying to only pass regions of proper rust code into it.

A lot of emacs packages for formatters has taken this approach, it's quite common. It should be easy to make rustfmt support it, I've just never raised the question.

blaenk:
Yeah. rustfmt would be in the best position to interpret and potentially fix start/end positions. i.e. if the start provided is in the middle of a parameter list, it's in the best position to determine if the start should be moved to the beginning of the list or after, etc.

nikomatsakis:
> While I currently can't think of a scenario where one would want to format just a specific region, the idea that that would be possible is very nice. Any chance we can do this? If not now, then after this has been merged? It seems like most of the work is already there.

FWIW, I use this capability in emacs all the time (that is, "indent-region"). Usually if I've just pasted some text. In principle of course if the rest of the file is formatted it doesn't make a difference I guess. Anyway, if rustfmt doesn't support it, oh well.

nikomatsakis:
This looks good to me but I only pretend to know Elisp. @MicahChalmer what do you think? :)

MicahChalmer:
Looks good to me too...I took the liberty of merging it.

nikomatsakis:
:dancer:

PR #123, opened 2016-02-25
Fix font-locking for Unicode escapes in character literals.

birkenfeld:
* There is only \u, no \U
* \u requires braces, but allows 1-6 hex digits


nikomatsakis:
Looks good! Thanks.

PR #129, opened 2016-03-04
Run the after-revert hook after rustfmt

### LINKED ISSUE #127: rustfmt causes indentation to get messed up
I haven't been able to narrow this down yet to a smaller buffer, but here is a consistent reproduction of a problem (for me at least). Start with [this input](https://gist.github.com/nikomatsakis/2ce0c51f1bcca16c1ec8) in some buffer. If you jump to the end and hit tab (M-x indent-for-tab-command) on the line `// need some logic for start state and eof`, then you should find that it stays put:

```
        ...
        // need some logic for start state and eof
    }
}
```

But if you then run `C-c C-f` (run rustfmt) and hit tab again, you will find that it jumps to the beginning of the line:

```
// need some logic for start state and eof
    }
}
```

If you close the buffer and re-open, things work fine again.

I encounter this quite regularly -- but my attempts to delete text out of the buffer often caused the issue to go away, so I'm leaving it as is.

cc @MicahChalmer this seems related to a previous issue we had with losing sync, iirc?

MicahChalmer:
Fix #127

(No tests, but there's no tests for the rest of the rustfmt feature anyway.  Such tests would either have to depend on rustfmt being installed, or fake it out in some annoying way anyway.)

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1247,6 +1247,12 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
     (rust--format-call (current-buffer))
     (goto-char cur-point)
     (set-window-start (selected-window) cur-win-start))
+
+  ;; Issue #127: Running this on a buffer acts like a revert, and could cause
+  ;; the fontification to get out of sync.  Call the same hook to ensure it is
+  ;; restored.
+  (rust--after-revert-hook)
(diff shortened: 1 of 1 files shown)


nikomatsakis:
makes sense

PR #131, opened 2016-03-05
Use modern properly-namespaced Common Lisp control macros (closes #130)

### LINKED ISSUE #130: When byte-compiled, gives "Error during redisplay: (jit-lock-function XXX) signaled (void-function block)"
I'm experiencing this very strange problem. When I run rust-mode.el everything is fine, but when I run it in byte-complied form as rust-mode.elc, I get this error lots of times. There are two ways to work around it:

- add `(require 'cl)` to my `init.el` before I load rust-mode, as mentioned in racer-rust/emacs-racer#14
- in `rust-mode.el`, change uses of `block` and `return` to `cl-block` and `cl-return` and require `cl-lib` not `cl` as prescribed in the [Emacs CL manual](https://www.gnu.org/software/emacs/manual/html_node/cl/Overview.html).

What is confusing me however, is that everything I'm reading about [eval-when-compile](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-During-Compile.html) which rust-mode is already using, suggests that I should be having the *opposite* problem, i.e. the problem should exist only when I'm *not* using the byte-compiled version. But I'm not an emacs expert, and I could just be interpreting these documents incorrectly.

infinity0:
Incompatible with Emacs 24.2 and earlier, but that's several years old now.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -14,7 +14,7 @@
 ;;; Code:
 
 (eval-when-compile (require 'rx)
-                   (require 'cl)
+                   (require 'cl-lib)
                    (require 'compile)
                    (require 'url-vars))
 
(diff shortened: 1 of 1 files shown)


kraai:
The comment on line 28 indicates that the following code is for compatibility with Emacs < 24.3.  If we remove compatibility with Emacs 24.2 and earlier, should this code also be removed?

infinity0:
I should mention that this is just a tentative suggestion. We should probably first figure out exactly why #130 is happening. Hopefully another person can reproduce this and shed some light...

tromey:
For a simple use like this, it's best not to get into the `cl` -vs `cl-lib` morass, and instead just use plain old `catch` and `throw`.

tomjakubowski:
> The comment on line 28 indicates that the following code is for compatibility with Emacs < 24.3. If we remove compatibility with Emacs 24.2 and earlier, should this code also be removed?

The Travis builds still test against Emacs 23, so this seems like a rather serious question to answer!

tromey:
This can be closed now.

MicahChalmer:
Merged #133 as an alternative instead.

infinity0:
Thanks!

PR #132, opened 2016-03-05
Add rust-promote-module-into-dir

### LINKED ISSUE #128: command to convert `foo.rs` into `foo/mod.rs` [feature-request]
It happens to me pretty regularly that I editing `foo.rs` and then decide to "promote" this into a directory-based module, usually so I can add a `foo/test.rs`, but sometimes for other reasons. I'd love a command to do this automatically!

tomjakubowski:
Lacking tests and completely untested on Windows!

closes #128

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1402,6 +1402,28 @@ See `compilation-error-regexp-alist' for help on their format.")
   (interactive)
   (rust-playpen-region (point-min) (point-max)))
 
+(defun rust-promote-module-into-dir ()
+  "Promote the module file visited by the current buffer into its own directory.
+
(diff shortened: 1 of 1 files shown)


tomjakubowski:
cc @nikomatsakis

tomjakubowski:
Ah, d'oh... no `file-name-base` on Emacs 23. Pushing a fix.

nikomatsakis:
ooh, neat, I should give this a spin. I have no idea how we would test it. :)

nikomatsakis:
OK so I tried this a few times and...it seems to work for me. :) I say let's land it! We can always fix whatever obscure bugs crop up later. Thanks!

PR #133, opened 2016-03-06
Use catch and throw instead of block and return

### LINKED ISSUE #130: When byte-compiled, gives "Error during redisplay: (jit-lock-function XXX) signaled (void-function block)"

kraai:
Fixes #130 (hopefully)

This also supersedes #131.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -14,7 +14,6 @@
 ;;; Code:
 
 (eval-when-compile (require 'rx)
-                   (require 'cl)
                    (require 'compile)
                    (require 'url-vars))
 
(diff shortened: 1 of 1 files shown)


nikomatsakis:
cc @MicahChalmer this is way past my level of skill with elisp to review :)

tromey:
@nikomatsakis fwiw I think this is fine.  The issue is that the `cl-` versions (`block` and `return`) have different names in different versions of Emacs, and they are just macro wrappers for `catch` and `throw` anyway.

MicahChalmer:
Yup, makes sense to me too.

PR #134, opened 2016-03-07
Allow disabling rustfmt with a dir-local/file-local variable

Stebalien:


Stebalien:
Not the way to do this...

PR #135, opened 2016-03-07
Allow disabling rustfmt with a dir-local/file-local variable

Stebalien:


rust-mode.el
```diff
@@ -1317,9 +1318,11 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (setq-local parse-sexp-lookup-properties t)
   (setq-local electric-pair-inhibit-predicate 'rust-electric-pair-inhibit-predicate-wrap)
   (add-hook 'after-revert-hook 'rust--after-revert-hook 'LOCAL)
```
Stebalien: Unrelated but this looks wrong. Shouldn't this end with `nil t` (or `nil 'LOCAL`), not just `'LOCAL`?

tomjakubowski: Yeah, that sounds right to me.

rust-mode.el
```diff
@@ -186,6 +186,7 @@ function or trait.  When nil, where will be aligned with fn or trait."
   "Format future rust buffers before saving using rustfmt."
   :type 'boolean
   :safe #'booleanp)
+(put 'rust-format-on-save 'safe-local-variable 'booleanp)
```
mrBliss: Isn't this line redundant given that the line above expresses exactly the same thing (in a nicer way)?

`defcustom`'s documentation says:

```
...
:safe	Set SYMBOL's `safe-local-variable' property to VALUE.
        See Info node `(elisp) File Local Variables'.
...
```

Stebalien: Didn't know that... thanks!

Stebalien:
I ended up just tacking on the second bug fix. I can split it into two pull requests if you'd like but I did it this way so I wouldn't forget.

rust-mode.el
MicahChalmer: Yup - that was a mistake.  Thanks for catching it.  Fortunately the hook would usually be fairly harmless when run in other modes...but much better if it's local as intended.

PR #136, opened 2016-03-07
Re-indent on }

Stebalien:
Before:

    fn test() {
        // Do something
        }▎

After:

    fn test() {
        // Do something
    }▎


tromey:
FWIW this patch seems reasonable to me.

nikomatsakis:
LGTM.

PR #138, opened 2016-03-08
Document how to install on Debian

kraai:
Once a new version of Debian is released (probably in mid-2017), I'll update this to document how to install rust-mode in Debian stable as well.


nikomatsakis:
thanks.

PR #141, opened 2016-03-08
set derived-mode-parent for rust-mode

### LINKED ISSUE #139: Inherit from prog-mode?
Is there a reason why `rust-mode` doesn't [derive from `prog-mode`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Major-Modes.html)? If not could you change it to do so?

I'm asking because it's caused a minor issue for someone trying to use [electric-operator](https://github.com/davidshepherd7/electric-operator/issues/32) with `rust-mode`, nothing that can't be worked around if necessary though.

tromey:
When prog-mode is defined, set derived-mode-parent for rust-mode to
prog-mode.  This helps derived-mode-p work properly.  This is only
needed due to the rust-parent-mode compatibility hack; when that goes
away, this can as well.

Fixes #139.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -2520,6 +2520,12 @@ Fontification needs to include this whole string or none of it.
     )
   )
 
+(when (fboundp 'prog-mode)
+  (ert-deftest rust-derived-mode ()
+    (with-temp-buffer
+      (rust-mode)
+      (should (derived-mode-p 'prog-mode)))))
+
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Seems reasonable, but honestly I have no idea what is going on here. :)

MicahChalmer:
Closing as no longer needed after #142 was merged.

PR #142, opened 2016-03-09
remove emacs 23 support

### LINKED ISSUE #137: How long should rust-mode keep supporting Emacs 23?
Is anyone using `rust-mode` regularly with emacs 23?  (The last version of emacs 23 was released in January 2012...three and a half years before Rust 1.0 was released, for what it's worth.)

I'm starting to see signs of slight hiccups for people trying to enhance rust-mode.  Nothing hard to overcome yet, but it could get more annoying later.

When is it time to cut it off?  When emacs 25 comes out?  When someone writes a feature that really needs something only available in emacs 24 and above?  Sooner than that?

### LINKED ISSUE #139: Inherit from prog-mode?

tromey:
This removes the various Emacs 23 compatibility hacks.

Fixes #137.
Fixes #139.
This also obsoletes PR #141.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,10 @@ language: generic
 
 env:
   matrix:
-    - EMACS=emacs23
     - EMACS=emacs24
     - EMACS=emacs-snapshot
 
 install:
-  - if [ "$EMACS" = 'emacs23' ]; then
-       sudo apt-get -qq update &&
-       sudo apt-get -qq -f install &&
-       sudo apt-get -qq install emacs23-gtk emacs23-el;
-    fi
   - if [ "$EMACS" = 'emacs24' ]; then
        sudo add-apt-repository -y ppa:cassou/emacs &&
        sudo apt-get -qq update &&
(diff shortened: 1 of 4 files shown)


tromey:
Looks like the comment was slightly incorrect and Emacs 24's ERT doesn't have `skip-unless`.  I'll fix it up a bit later.

MicahChalmer:
Could you also take out the bits in the README that have instructions on how to use it with emacs 23 (one on downloading `ert.el`, and one on getting `package.el`) and add a note there that rust-mode now requires emacs 24.1 or later?

Other than that, this looks good to me--assuming (as per the discussion in #137) that now is the time to drop emacs 23 support.

nikomatsakis:
On Wed, Mar 09, 2016 at 05:49:40PM -0800, Micah Chalmer wrote:
> Other than that, this looks good to me--assuming (as per the discussion in #137) that now is the time to drop emacs 23 support.

+1

MicahChalmer:
Nobody's objected, I don't think anyone will object...I'm merging it.

PR #148, opened 2016-04-05
Fill regular block comments correctly

bmastenbrook:
Right now fill-paragraph fills non-rustdoc block comments of the following form:

```
/* Four score and seven years ago our fathers brought forth on this continent a new nation, conceived in liberty, and dedicated to the proposition that all men are created equal. */
```

Like so: 
```
/* Four score and seven years ago our fathers brought forth on this
/* continent a new nation, conceived in liberty, and dedicated to the
/* proposition that all men are created equal. */
```

This doesn't work very well because Rust comments nest properly. This patch makes these comments format like so:

```
/* Four score and seven years ago our fathers brought forth on this
 * continent a new nation, conceived in liberty, and dedicated to the
 * proposition that all men are created equal. */
```

Which works a whole lot better. :-)

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -164,6 +164,15 @@ Also, the result should be the same regardless of whether the code is at the beg
      //
      // This is the second really really really really really really long paragraph" 1 89))
 
+(ert-deftest fill-paragraph-multi-line-style-comment ()
+  (test-fill-paragraph
+   "/* This is a very very very very very very very very long string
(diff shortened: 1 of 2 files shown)


bmastenbrook:
I've fixed the failing tests and added a test for the fix.

bmastenbrook:
Third time's the charm.

nikomatsakis:
Thanks!

PR #153, opened 2016-04-24
Fix #151

### LINKED ISSUE #151: indentation misaligned due to `where` in comment
This example indents as shown:

```rust
/// - there must not exist an edge U->V in the graph where:
#[derive(Clone, PartialEq, Eq)]
    pub struct Region {
    entry: BasicBlockIndex,
    leaves: BTreeMap<BasicBlockIndex, usize>,
}
```

But I would expect:

```rust
/// - there must not exist an edge U->V in the graph where:
#[derive(Clone, PartialEq, Eq)]
pub struct Region { // <-- this should be flush with left margin!
    entry: BasicBlockIndex,
    leaves: BTreeMap<BasicBlockIndex, usize>,
}
```

Presumably the angle-bracket matching code is going a bit awry here.

cc @MicahChalmer

mrBliss:
When looking for a `where` in
`rust-rewind-to-beginning-of-current-level-expr`, check that it is not
part of a string or comment.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -623,6 +623,17 @@ impl<'a, K, Q: ?Sized, V, S> Index<&'a Q> for HashMap<K, V, S>
 }
 "))
 
+(ert-deftest indent-align-where-in-comment ()
+  (test-indent
+   "/// - there must not exist an edge U->V in the graph where:
+#[derive(Clone, PartialEq, Eq)]
+pub struct Region { // <-- this should be flush with left margin!
+    entry: BasicBlockIndex,
+    leaves: BTreeMap<BasicBlockIndex, usize>,
+}
+"))
+
+
 (ert-deftest indent-square-bracket-alignment ()
   (test-indent
    "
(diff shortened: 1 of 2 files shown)


MicahChalmer:
Even with this patch, there are still places in the code that look for where clauses and don't check for comments, so there are still edge cases in which the word "where" in a comment can affect the indentation.  For instance, with or without this PR, rust-mode will indent the following code as follows:

````rust
fn foo<F,G>(f:F, g:G)
    where F:Send,
// where
    G:Sized
{
    
}
````

but if you change the "where" in the comment to another word, the `G:Sized` is aligned with the `F:Send` (which I believe is correct.)  If you make the comment a a where clause with a type annotation in it, rust-mode starts to align the actual code with the comment:

````rust
fn foo<F,G>(f:F, g:G)
    where F:Send,
// where      F:ThisIsNotActualCode,
              G:Sized
{
    
}
````

mrBliss:
You're right. I'll update all places where we're looking for `"where"` to use (the existing) `rust-in-str-or-cmnt`.

PR #154, opened 2016-05-04
add code to handle new-style rustc errors

nikomatsakis:
These errors are available on nightly builds (or will be soon), but only (at the moment) when enabled via environment variable. They will become the default at some point in the future. In this commit we match on the `-->`, but after that we have to scroll the compilation window to make the error visible. 

(One shortcoming of this commit is that the resulting colors could be better: emacs highlights the filename and line-number, but not the error message itself. I consider the result suboptimal. I considered adding more logic here to add a "bold" attribute to the error message lines, but haven't had time to try it out.)

r? @MicahChalmer

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1381,6 +1381,16 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   "Specifications for matching errors in rustc invocations.
 See `compilation-error-regexp-alist' for help on their format.")
 
+(defvar rustc-new-compilation-regexps
+  (let ((file "\\([^\n]+\\)")
+        (start-line "\\([0-9]+\\)")
+        (start-col  "\\([0-9]+\\)"))
+    (let ((re (concat "^ *--> " file ":" start-line ":" start-col ; --> 1:2:3
+                      )))
+      (cons re '(1 2 3))))
+  "Specifications for matching errors in rustc invocations (new style).
+See `compilation-error-regexp-alist' for help on their format.")
+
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -1391,12 +1401,37 @@ See `compilation-error-regexp-alist' for help on their format.")
...
+              (looking-at " *-->"))
+        (let ((start-of-error
+               (save-excursion
+                 (beginning-of-line)
+                 (while (not (looking-at "^[a-z]+:"))
+                   (forward-line -1))
+                 (point))))
+          (set-window-start (selected-window) start-of-error))))))
+
+(add-hook 'next-error-hook 'rustc-scroll-down-after-next-error)
```
nikomatsakis: ah, I should probably move this into `eval-after-load 'compile`

MicahChalmer:
Interesting discussion happening on https://internals.rust-lang.org/t/editor-compatibility-and-the-new-error-format/3435/21.  I would hope that somehow there's a way to get it to work without the hack...but since the new style errors are in rustc nightly, better to put this in for now.  The hack can be removed if it becomes unnecessary (as in your suggestion on rust-lang/rust#32486.)

The hook added here will scroll the compilation window so that the start of the error message is at the very top of it, even if that wasn't necessary to bring the error message into view.  That is, it will scroll away anything above the error, even if the error was already visible before the `next-error` call.  That may not be a bad thing (if the error is going to print out a big block of code, I'd even say it's a good thing) but it's different than the normal behavior for other compilations.

PR #157, opened 2016-05-14
Improve imenu support

### LINKED ISSUE #93: everything shows up as Function in imenu
Is this expected or is it something on my end? Everything except for impls shows up as "Function" on my end, is that what is meant by "Imenu will show all the enums, structs, etc. at the same level." in the source comment? Is there a reason we can't have different headings like "Struct" and "Enum" etc.?

### LINKED ISSUE #94: imenu regexps are confused by comments
Given the code

```rust

// struct X

struct Y {
    A: i64,
}

 struct Z {
    A: i64,
}

"
struct Q {
    A: i64,
}
"

fn foo() {}

// fn bar() {}
```

imenu offers X, Y, Z, foo and bar. I'm not sure why this occurs, since `imenu--generic-function` tries hard to avoid strings and comments (unless you modify `imenu-generic-skip-comments-and-strings`).

fmdkdd:
This PR does three things to the imenu support of rust-mode:

- Fix #94 (avoid picking up items in single-line comments)
- Fix #93 (all items show up as top-level in imenu)
- Adds macro_rules names to imenu

#### Fix #94
#94 is definitely a bug.  Imenu picked up item definitions in single-line comments, but not multi-line comments apparently.

I've added a regexp derived from `rust-re-item-def` that strictly matches from the beginning of the line to avoid.  Since items can have a `pub` keyword for visibility in front of them, I match that optionally.

I did not modify `rust-re-item-def` since it was used for font-locking as well as imenu.  Though it should work there as well.

#### Fix #93
#93 seemed like a deliberate choice, since the comment for the `rust-imenu-generic-expression` is coherent with the code.

I think it's more useful to have separate categories, since I can immediately distinguish between an `impl` and a `struct` of the same name.  But I only use imenu through Helm, so it might have downsides for other use cases.

You can cherry-pick, or I can update the PR if you'd rather have every definition (except impls) at the top-level.

#### Add macro_rules
I love macros!  rust-mode did not add them to imenu previously.  But I find it very useful to jump to them in large files.

This is just a matter of adding `macro_rules!` to the items we map over in defining `rust-imenu-generic-expression`.  But this doesn't work yet.  I suppose this is because of the `!`, which is not picked up by the regexp that uses word boundaries.  Adding `!` to the imenu syntax table let us reuse the regexp.

#### Correctness
I've tested the PR to work locally.  It doesn't break any ERT test (though there may not be any imenu-specific tests).

My understanding of all that stuff is quite superficial, so there might be downsides to the way I've approached it.  I can update the PR as needed per review.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1195,16 +1195,15 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
 ;;; Imenu support
 (defvar rust-imenu-generic-expression
   (append (mapcar #'(lambda (x)
-                      (list nil (rust-re-item-def-imenu x) 1))
-                  '("enum" "struct" "type" "mod" "fn" "trait" "macro_rules!"))
-          `(("Impl" ,(rust-re-item-def-imenu "impl") 1)))
+                      (list (capitalize x) (rust-re-item-def-imenu x) 1))
+                  '("enum" "struct" "type" "mod" "fn" "trait" "impl"))
+          `(("Macro" ,(rust-re-item-def-imenu "macro_rules!") 1)))
   "Value for `imenu-generic-expression' in Rust mode.
 
-Create a flat index of the item definitions in a Rust file.
+Create a hierarchical index of the item definitions in a Rust file.
 
-Imenu will show all the enums, structs, etc. at the same level.
-Implementations will be shown under the `Impl` subheading.  Use
-idomenu (imenu with `ido-mode') for best mileage.")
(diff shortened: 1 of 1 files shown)


blaenk:
This is great! I'm delighted to see someone saw #93 haha. I'd love it if the fix for #93 is merged even if it's opt-in, it would be very nice!

nikomatsakis:
Looks good to me. Thanks!

PR #159, opened 2016-07-15
[master] Copy buffer only when format changes.

hotpxl:
In formatting code, only copies when `rustfmt` actually changed something. This will save some meaningless "changed buffer" signs even if there is no change.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1258,8 +1258,10 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
     (erase-buffer)
     (insert-buffer-substring buf)
     (if (zerop (call-process-region (point-min) (point-max) rust-rustfmt-bin t t nil))
-        (progn (copy-to-buffer buf (point-min) (point-max))
-               (kill-buffer))
+        (progn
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -1258,8 +1258,10 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
     (erase-buffer)
     (insert-buffer-substring buf)
     (if (zerop (call-process-region (point-min) (point-max) rust-rustfmt-bin t t nil))
-        (progn (copy-to-buffer buf (point-min) (point-max))
-               (kill-buffer))
+        (progn
+          (if (not (string= (buffer-string) (with-current-buffer buf (buffer-string))))
+              (progn (copy-to-buffer buf (point-min) (point-max))))
```
kraai: I don't think it's necessary to use progn here anymore, since there's only one form inside the if's then-part.

kraai:
@hotpxl Thanks!

hotpxl:
Hi. Just want to check on the status. Is there anything else I should do?

nikomatsakis:
LGTM :)

PR #161, opened 2016-07-19
Fix #160

### LINKED ISSUE #160: parentheses all turn red when format codes
When I'm using spacemacs, after formated the code (C-c C-f), the parenthese turn red. I'm sure the syntax of code is correct.

If I type a M-x rust-mode (or any other mode, the red things will disappear).

I did not make any special configuration, just the default rust layer:
https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Blang/rust

on the other hand, when I use my own configuration without spacemacs, just "racer, ac-racer, company-racer, toml-mode", there is no such issue.

not sure it's a spacemacs rust layer issue or a rust-mode issue, or something wrong between them.

there are some talk in the below link:
https://github.com/fbergroth/emacs-rustfmt/issues/9

parenthese_turn_red

I've also post an issue in spacemacs repo:
https://github.com/syl20bnr/spacemacs/issues/6617
anybody could help on this?
Thanks!

Fanael:
Use `font-lock-fontify-region` on the whole buffer to actually force refontification in Emacs 24, which does not support `font-lock-ensure`.

`font-lock-fontify-buffer` is actually closer to `font-lock-flush`, so it's not the right function to use in `rust--after-revert-hook`.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1359,13 +1359,13 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 ;; Issue #104: When reverting the buffer, make sure all fontification is redone
 ;; so that we don't end up missing a non-angle-bracket '<' or '>' character.
 (defun rust--after-revert-hook ()
-  (let
-      ;; Newer emacs versions (25 and later) make `font-lock-fontify-buffer'
-      ;; interactive-only, and want lisp code to call `font-lock-flush' or
-      ;; `font-lock-ensure'.  But those don't exist in emacs 24 and earlier.
-      ((font-lock-ensure-fn (if (fboundp 'font-lock-ensure) 'font-lock-ensure 'font-lock-fontify-buffer)))
-    (funcall font-lock-ensure-fn))
-  )
+  ;; In Emacs 25 and later, the preferred method to force fontification is
+  ;; to use `font-lock-ensure', which doesn't exist in Emacs 24 and earlier.
+  ;; If it's not available, fall back to calling `font-lock-fontify-region'
+  ;; on the whole buffer.
+  (if (fboundp 'font-lock-ensure)
+      (font-lock-ensure)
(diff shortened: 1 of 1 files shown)


Fanael:
Ultimately, bugs like #104 and #160 stem from the way `rust-mode` tries to determine whether `<` and `>` are operators or angle brackets. `c++-mode` manages to get it right without getting `syntax-ppss` horribly confused, so you can try contacting its maintainers for advice.

nikomatsakis:
> Ultimately, bugs like #104 and #160 stem from the way rust-mode tries to determine whether `<` and `>` are operators or angle brackets. `c++-mode` manages to get it right without getting syntax-ppss horribly confused, so you can try contacting its maintainers for advice.

@MicahChalmer is the one who knows that part of the system best. All I can say is that now that we have it working properly *most* of the time, I would never want to give it up. =)

nikomatsakis:
This patch seems good to me, though the subtle distinctions between `font-lock-fontify-buffer` and `font-lock-fontify-region` are over my head. =)

liuchong:
which version of emacs24 you are using has the "font-lock-fontify-region"? I'm still have the red-parentheses problem.
anyway, I will upgrade my emacs to 25.

Fanael:
> the subtle distinctions between `font-lock-fontify-buffer` and `font-lock-fontify-region` are over my head

They both basically do nothing but call `font-lock-fontify-{buffer,region}-function`. The difference is that, in the default configuration, `font-lock-fontify-region-function` is `font-lock-default-fontify-region`, which *immediately* fontifies the passed region (i.e. behaves like `font-lock-ensure`); while `font-lock-fontify-buffer-function` is `jit-lock-refontify`, which merely marks the whole buffer as not yet fontified (i.e. behaves like `font-lock-flush`).

Issue #162, opened 2016-08-02
rust-format-buffer does not preserve point position


Wilfred:
Suppose I have the following code, where | is the point:

``` rust
fn foo(foo_bar_baz1: i32, foo_bar_baz2: i32, foo_bar_baz3: i32, foo_bar_baz4: i32, foo_bar_baz5: i32, foo_bar_baz6: i32) {

}

fn bar() {|}
```

Calling `M-x rust-format-buffer` naively moves to the same position after formatting, putting point in the arguments to `foo`. It then calls `rust--after-revert-hook`, which moves point to the start of the buffer.

Ideally, we'd use rustfmt's `--write-mode=diff` so emacs-rustfmt could just apply the diff. Emacs' markers are smart enough to move if text is inserted before them. This is how [gofmt is integrated with Emacs](https://github.com/dominikh/go-mode.el/blob/5d53a13bd193653728e74102c81aa931b780c9a9/go-mode.el#L934).

A simple partial fix would be to move to the same line number and column number as before -- this would handle more cases correctly.

tromey:
One possible fix might be to count non-whitespace characters before point, and then skip that many non-whitespace characters after restoring the buffer.  The idea here is that presumably the reformat operation affects whitespace but nothing else.

tromey:
> The idea here is that presumably the reformat operation affects whitespace but nothing else.

This turns out to be incorrect; rustfmt can change comment styles.

Dushistov:
@tromey 

I also have problem with this, I setted `rust-format-on-save` to true, and this hurted me.
Time to time my cursor  jump to begin of buffer. So I have to turn off  `rust-format-on-save` for now, and run `M-x cargo-process-fmt` time to time by hands.

May be ask `rustfmt` team for feature: give to `rustfmt` position in file, and after formating `rustfmt` return new position? `rustfmt` can works with AST tree, and calculate cursor position against this AST tree, and recalculate based on changes in this AST tree.

rofrol:
This is my solution based on https://github.com/rust-lang/rust-mode/pull/269#issuecomment-406780226

```elisp
;; -*- lexical-binding: t; -*-

(use-package rust-mode
    :mode ("\\.lalrpop\\'" . elm-mode)
    :straight t
    :config
    (setq rust-format-on-save nil)
    (defun rust-mode-format-buffer ()
      "format rust code without jumping to beginning when doing undo"
      (interactive)
      (shell-command
       (format "rustfmt %s"
[...]

PR #163, opened 2016-08-03
Correctly restore point position after running rustfmt

Wilfred:
The previous approach simply moved to the same character offset. This is
unlikely to preserve the position of point, as rustfmt often changes
whitespace, changing the number of characters before point.

Instead, we go back to the line number and column number we were on
before. Provided that rustfmt has not radically changed the number of
lines, this will typically put point back to its previous position, or
at least close.

Improves, but doesn't completely solve, issue #162.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1365,9 +1365,10 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   ;; to use `font-lock-ensure', which doesn't exist in Emacs 24 and earlier.
   ;; If it's not available, fall back to calling `font-lock-fontify-region'
   ;; on the whole buffer.
-  (if (fboundp 'font-lock-ensure)
-      (font-lock-ensure)
-    (font-lock-fontify-region (point-min) (point-max))))
+  (save-excursion
+    (if (fboundp 'font-lock-ensure)
+        (font-lock-ensure)
+      (font-lock-fontify-region (point-min) (point-max)))))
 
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -1365,9 +1371,10 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   ;; to use `font-lock-ensure', which doesn't exist in Emacs 24 and earlier.
   ;; If it's not available, fall back to calling `font-lock-fontify-region'
   ;; on the whole buffer.
-  (if (fboundp 'font-lock-ensure)
-      (font-lock-ensure)
-    (font-lock-fontify-region (point-min) (point-max))))
+  (save-excursion
+    (if (fboundp 'font-lock-ensure)
+        (font-lock-ensure)
+      (font-lock-fontify-region (point-min) (point-max)))))
```
nikomatsakis: how is this change related?

Wilfred: `rust-format-buffer` calls this function after repositioning point. If this function moves point, then our repositioning of point doesn't do us any good.

Does that make sense? This is a separate commit, and I'm happy to amend the commit message if I can make it clearer.

Wilfred:
@nikomatsakis ping, are you happy with this? I think it's a worthwhile improvement. Are there any changes you'd like me to make?

nikomatsakis:
Ah, sorry, I agree!

PR #165, opened 2016-08-15
fix regex to account for `[E123]`

nikomatsakis:
The old regex was fine when it was written, but the style changed since then.


nikomatsakis:
I'm going to land this because it seems pretty important to have. @MicahChalmer if you want to review post-facto, sounds good to me.

PR #169, opened 2016-09-09
Fix #168: use while in rust-rewind-irrelevant

### LINKED ISSUE #168: Stack overflow hitting return in long document or long comment
If I have a large document, presumably with a lot of comments and I hit return after a certain number of bytes, I get the following error:

```
Debugger entered--Lisp error: (error "Lisp nesting exceeds `max-lisp-eval-depth'")
  #[0 blah blah blah lots of binary strings blah blah])
  syntax-ppss()
  rust-in-str-or-cmnt()
  rust-rewind-irrelevant() [... 589 times ...]
  rust-mode-indent-line()
  indent-according-to-mode()
  electric-indent-post-self-insert-function()
  self-insert-command(1)
  newline(nil 1)
  call-interactively(newline nil nil)
  command-execute(newline)
```

Worse, it leaves me in a very broken state and all typing appears "backwards". What is actually happening is some half-functioning macro is prepended to post-self-insert-hook that effectively runs move-beginning-of-line on every character, causing you to type on the front of the current line. To fix that, I have to `M-:` and paste in:

```
(setq post-self-insert-hook '(electric-indent-post-self-insert-function blink-paren-post-self-insert-function))
```

which removes the bad hook.

The problem appears to be:

```
(defun rust-rewind-irrelevant ()
  (let ((starting (point)))
    (skip-chars-backward "[:space:]\n")
    (if (rust-looking-back-str "*/") (backward-char))
    (if (rust-in-str-or-cmnt)
        (rust-rewind-past-str-cmnt))
    (if (/= starting (point))
        (rust-rewind-irrelevant))))
```

But elisp doesn't have tail-call optimization, so it blows out and doesn't clean up after itself. This function can be rewritten with a named let loop or something... or the calling function should have an unwind handler on it.

Oh... and large is ~665 lines or ~16000 chars. Not sure which of the two is relevant, but given the above stack trace, it is probably the number of lines.

mrBliss:
Rewrite the recursive function `rust-rewind-irrelevant`, which causes a
stack overflow in #168, using a `while` loop.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1162,6 +1162,23 @@ All positions are position symbols found in `rust-test-positions-alist'."
    'nonblank-line-indented-already-middle-target
    #'indent-for-tab-command))
 
+(ert-deftest no-stack-overflow-in-rust-rewind-irrelevant ()
+  (with-temp-buffer
+    (rust-mode)
+    (insert "fn main() {\n    let x = 1;")
+    ;; Insert 150 separate comments on the same line
+    (dotimes (i 150)
+      (insert "/* foo */ "))
+    ;; Rewinding from the last commment to the end of the let needs at least
+    ;; 150 iterations, but if we limit the stack depth to 100 (this appears to
+    ;; be some minimum), a recursive function would overflow, throwing an
+    ;; error.
+    (let ((max-lisp-eval-depth 100))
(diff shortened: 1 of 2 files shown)


mrBliss:
I'll try to add a test for this.

PR #173, opened 2016-10-02
Add coloring of type-inferred constants and variables

nordlow:


pnkfelix:
Um, sorry: What is a type-inferred constant/variable? Just one that does not have an explicit type on its binding?  (But don't all constant definitions require an explicit type?)

nikomatsakis:
I think "constant" is being used to refer to a shared binding (e.g., `x` in
`let x = 3`). I agree it's not a term I would use, particularly since the
binding may in fact include mutable data:

```rust
let x = Cell::new(0);
x.set(1);
println!("{}", x.get()); // prints 1, not 0
```

nikomatsakis:
@nordlow did you see the conflicts and travis failures? Maybe you can rebase and take a look? 

Sorry that this PR has sat without feedback for so long!

nordlow:
Perhaps both let and let mut "variables" should use `font-lock-variable-name-face`. It might be that font-lock-constant-face is reserved for compile-time constants.

nikomatsakis:
@nordlow I am no expert, but that sounds correct

nordlow:
Rebased.

nikomatsakis:
@nordlow personally I would color it the same

nordlow:
Why do the checks fail?

rust-mode.el
```diff
@@ -616,6 +616,12 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
      ;; Field names like `foo:`, highlight excluding the :
      (,(concat (rust-re-grab rust-re-ident) ":[^:]") 1 font-lock-variable-name-face)
 
+     ;; Type-inferred constant
+     (,(concat "\\_<\\(?:let\\|ref\\)\\s-+" (rust-re-grab rust-re-ident) "\\_>") 1 font-lock-constant-face)
```
nikomatsakis: I do not think of this as a constant. It is a variable, just one that doesn't permit mutable borrows (note that interior mutation can still occur, e.g. with a `Cell`). Can you change to `font-lock-variable-name-face` and group with the regex below?

nikomatsakis:
@nordlow 

> Why do the checks fail?

It looks like the existing tests include some type-inferred variables, so you'll have to update the expected result to include your changes.

Hi-Angel:
Hello, any movement? Incidentally, I appeared there just because I was going to make the same changes, and then found this PR. I need the `font-lock-variable-name-face` *(and may be the constant one?)* to make it working with [color-identifiers](https://github.com/ankurdave/color-identifiers-mode) mode.

Hi-Angel:
Awesome, with this patch adding `rust-mode` to the do-list of the `color-identifiers:cc-mode-get-declarations` makes it working!

@nordlow please can you address the @nikomatsakis's request so they'd accept the PR, and I could make `color-identifiers` working with non-patched rust-mode?

Aankhen:
I checked this out to see what updating it would take and it was a simple fix. In case @nordlow doesn’t have time, I made the change @nikomatsakis asked for, updated the existing tests, and added a couple for the new functionality over at [my repo](https://github.com/Aankhen/rust-mode/tree/add-var-colours). I’m not sure how to proceed from here though.

Hi-Angel:
@Aankhen given the OP have had activity twice since, but still did not answer, I think it's ok if you pull an overriding request.

Aankhen:
Makes sense, @Hi-Angel. I’ve opened #225.

tromey:
I'm closing this because it's been superseded by another PR.

Issue #170, opened 2016-09-12
Rust 1.12 - new error format enabled


jonathandturner:
I just wanted to send out a friendly reminder that we're 17 days away from the release of Rust 1.12.  With Rust 1.12, we'll be switching to the new error format (and removing the old format).  There's also a standard JSON error format you can use.

I know @nikomatsakis has been watching this closely in emacs, since he and @wycats worked with me on the new format :)

Related: https://github.com/rust-lang/rust-mode/issues/156

Issue #156, opened 2016-05-14
New error format: all diagnostics are now errors


birkenfeld:
With the old error format, rust-mode was able to recognize errors, warnings and other info messages. The result was that compilation-mode could use different faces to distinguish between messages visually (`compilation-error`, `compilation-warning`, `compilation-info`).

This is now gone; all diagnostics are categorized as errors and highlighted as such (or rather, their position is). Since notes/help do no longer have individual file:line, it's not particularly serious, but unfortunate.

nikomatsakis:
Ah yes, I meant to address this by changing the new error format generating code.

qdot:
+1 on this since the new error format landed in nightly but compilation mode highlighting hasn't been updated yet.

blaenk:
Just wanted to state another reason for fixing this. Emacs commit https://github.com/emacs-mirror/emacs/commit/846870e508021ee8d1099280b3f40fe108a34bf0 added automatic support for counting each type of error, so that for example, the count of each can be displayed on the mode-line.

I added this information to my mode-line but indeed all errors regardless of type show up as errors (despite them all being warnings in my case, for example).

PR #174, opened 2016-10-05
Distinguish error/warning in compilation-mode

pnkfelix:
Use naive multiline regexp to match message type and file:line:col spec in one step.

Spawned off from https://github.com/rust-lang/rust-mode/pull/154#discussion_r81976657

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1406,9 +1406,17 @@ See `compilation-error-regexp-alist' for help on their format.")
   (let ((file "\\([^\n]+\\)")
         (start-line "\\([0-9]+\\)")
         (start-col  "\\([0-9]+\\)"))
-    (let ((re (concat "^ *--> " file ":" start-line ":" start-col ; --> 1:2:3
-                      )))
-      (cons re '(1 2 3))))
+    (let ((arrow-filespec (concat " *--> " file ":" start-line ":" start-col ; --> 1:2:3
+                                  )))
+      (let ((error-re (concat "^[Ee]rror\\[E[0-9]+\\]: [^\n]*\n" arrow-filespec))
+            (warn-re  (concat "^[Ww]arning: [^\n]*\n" arrow-filespec)))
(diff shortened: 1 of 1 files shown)


nikomatsakis:
So, as I wrote on IRC, I tried this but had problems because we wound sometimes flush buffers (on the rustc side) at inconvenient times. However, https://github.com/rust-lang/rust/pull/35975 seems to have fixed this behavior (at least on Mac/Unix), so I expect that multiline-regex will work now.

pnkfelix:
Hmm now that I understand the problem that @nikomatsakis described, it is possible that I am actually observing it within my emacs instance (just not as frequently as he was, at least in `rustc` builds).

I will look into changing the code so that it uses the multiline regexp first and falls back on the one line one if necessary. (But I suspect I'll continue to leave out the scrolling...)

pnkfelix:
cc #156

pnkfelix:
cc #170

nikomatsakis:
> I will look into changing the code so that it uses the multiline regexp first and falls back on the one line one if necessary. (But I suspect I'll continue to leave out the scrolling...)

Hmm, without the scrolling it is pretty annoying to use the one line regex, at least if you follow my workflow of doing `M-x next-error` a lot.

nikomatsakis:
@pnkfelix what's the story with your investigations into multi-line-regexp here? 

Have you been using this patch locally? I guess that if we have the fallback, this is no worse than the existing scenario, but it seems odd to me that it will be inconsistent in its ability to detect warnings. Still...

pnkfelix:
@nikomatsakis my experience is that using the multiline regexp hits a bug in compile.el, which I think you yourself had encountered?  I filed it against emacs and they confirmed the bug, number 25133 : https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25133

Doesn't help us resolve this in short term of course; but the basic answer is that for now the multi line regexp is a no go

nikomatsakis:
@pnkfelix maybe you can close this PR but file some notes in a suitable issue (do we have one already?)

good to know that they acknowledged it as a bug

PR #175, opened 2016-10-19
Recognize imenu items starting with "unsafe"

mrBliss:
Things like `unsafe fn foo(...` where left out of the imenu list.


nikomatsakis:
Thanks!

blaenk:
@mrBliss when you get a chance can you also do this for extern fns? See #188

Thanks for your work on this :)

PR #177, opened 2016-10-22
Handle comments when indenting method chains

mrBliss:
Previously, indentation went wrong when the comment on the line above looked like a method call/member access:

```rust
fn main() {
    // Lorem ipsum lorem ipsum lorem ipsum lorem.ipsum
                                            foo.bar()
}
```

With this patch:

```rust
fn main() {
    // Lorem ipsum lorem ipsum lorem ipsum lorem.ipsum
    foo.bar()
}
```

Also, a blank line or a comment broke method-chain indentation:

```rust
fn main() {
    something.a.do_it
    // A comment
        .aligned

        .more_alignment();
}
```

With this patch:

```rust
fn main() {
    something.a.do_it
    // A comment
               .aligned

               .more_alignment();
}
```

Note that comments interleaving a method chain are not aligned with the '.' of the method chain.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1599,6 +1599,19 @@ fn main() {
 "
    )))
 
+(ert-deftest indent-method-chains-look-over-comment ()
+  (let ((rust-indent-method-chain t)) (test-indent
+   "
+fn main() {
+    thing.a.do_it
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Great =)

Issue #87, opened 2015-07-21
Melpa stable


nlamirault:
could you please create a tag to use this package using melpa stable [1].
Thanks

[1] : http://stable.melpa.org/

MicahChalmer:
There isn't an obvious point to consider a release of rust-mode to be a "stable version" here.  PRs get merged, and if there are problems more PRs are merged to fix them.  If we added a tag, when would we be expected to add another?

Wilfred:
It makes it easier to track changes, and it's really helpful for packages that depend on rust-mode. Without MELPA stable there isn't really much versioning.

>  If we added a tag, when would we be expected to add another?

Monthly? After major bugfixes or feature additions?

The other nice advantage of versioning is that it allows communication of the severity of changes. If rust-mode went from 2.0 to 3.0 it suggests that more APIs may have changed.

Ketsuban:
The main wrinkle that I can see is the track system Rust uses—not everyone sticks to nightly. Perhaps the repository should have branches mirroring the Rust tracks with each one providing a different package to MELPA Stable, so that if I use Rust stable I can install `rust-mode-stable` in Emacs and know that it will always be the appropriate mode for the version of Rust I'm working against.

Wilfred:
I think all Rust users would want the latest version of rust-mode. The most recent changes have corrected highlighting, improved performance -- all things that that apply to all rust versions. Similarly, as a C programmer I wouldn't expect to have to update c-mode if I start a project in c99 instead of c89. 

When packages depend on rust-mode (only one so far), they might want to depend on a minimum version. E.g. `rust-playpen-region` was added a few months ago and if I explicitly need that function, I would want to specify a minimum version of rust-mode that includes that function.

nikomatsakis:
> I think all Rust users would want the latest version of rust-mode.

I agree. Also, Rust syntax is quite stable and certainly any changes will
be largely backwards compat, so I'd expect new versions of rust-mode to
work great on older rust code.

Skinney:
Any lingering thoughts on this? I'm more or less just waiting for rust-mode to hit melpa stable :P

Wilfred:
Who has the power to create tags? @nikomatsakis ?

nikomatsakis:
@Wilfred I can create tags, what tag do you want?

Wilfred:
How about 1.0?

nikomatsakis:
[...]

PR #178, opened 2016-10-27
Bump version

ryuslash:
Sorry if this is too forward, but it was suggested in #87 that someone who wants a (new) version to be released/tagged should just open a pull request, so that's what I'm doing. My hope is that you will create a git tag for this version so that it shows up in Melpa Stable, or at the very least get the release discussion going again. If you don't, then this pull request really has no meaning.

I thought I'd stay conservative with the new version number, as I just wanted to try some rust in Emacs today and found that rust-mode wasn't available on melpa stable and don't actually know anything about this project yet (in terms of goals, open issues, etc.). If this project is already in a state where you feel that all the initial goals of the project have been reached and no backward-incompatible changes will happen in the forseable future, then the version could of course be changed to 1.0.0 instead. All of this is assuming that you (will want to) adhere to semantic versioning, if you don't then any version number will do.

Creating a release isn't just about users wanting to install rust-mode while using melpa stable, but also about at least 2 packages (racer and cargo) which are in melpa stable, but can't be installed due to depending on an unavailable package (rust-mode).

In case you care about completeness, commits 4c1eab5 and 5afc8ab may also be prime candidates for tagging as they changed the versions to 0.2.0 and 0.1.0 respectively.

So whatever you do, _please_ tag a release.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1,6 +1,6 @@
 ;;; rust-mode.el --- A major emacs mode for editing Rust source code -*-lexical-binding: t-*-
 
-;; Version: 0.2.0
+;; Version: 0.3.0
 ;; Author: Mozilla
 ;; Url: https://github.com/rust-lang/rust-mode
 ;; Keywords: languages


pnkfelix:
reassigning @MicahChalmer ; you know a lot more about this than I do.

MicahChalmer:
Well I know as much as I said in #87.  It seems like there's a lot of people asking for this (somebody else opened #179 just after this was filed), and nobody objected to the proposal at the end of the discussion in #87.  So let's go with that.  I'll accept this as the first such pull request, and add the tag.

ryuslash:
Thank you! It's now on melpa stable: http://stable.melpa.org/#/rust-mode

brson:
Thanks @ryuslash !

PR #187, opened 2016-12-22
handle indirect buffers and multiple windows in rust-format-buffer

tspiteri:
Currently, if a buffer is shown in two windows, rust-format-buffer only tries to restore the cursor position in the current window. Also, if there are cloned indirect buffers (C-x 4 c), the position in other buffers is not restored either. This patch tries to restore position in all indirect buffers and all windows showing this buffer.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1287,17 +1287,60 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (unless (executable-find rust-rustfmt-bin)
     (error "Could not locate executable \"%s\"" rust-rustfmt-bin))
 
-  (let ((cur-line (line-number-at-pos))
-        (cur-column (current-column))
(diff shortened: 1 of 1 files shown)


tspiteri:
The second commit improves position recovery in rust-format-buffer by counting a number of specific keywords before the point, then counting lines and columns after the nearest matched word. This improves the situation of issue #162 considerably as the point does not move across functions/structs etc., and changes outside the current function do not affect us.

nikomatsakis:
Hmm I guess we don't have any tests for this, do we?

nikomatsakis:
(By "this" I mean rustfmt integration...)

nikomatsakis:
@tspiteri 

> The second commit improves position recovery in rust-format-buffer by counting a number of specific keywords before the point, then counting lines and columns after the nearest matched word.

clever!

PR #189, opened 2017-01-06
minor cleanups: https links and defcustom grouping

froydnj:
Just a few things I noticed while browsing through the code.


nikomatsakis:
Thanks!

PR #190, opened 2017-01-12
Recognize "extern" imenu items

### LINKED ISSUE #188: imenu: unsafe and extern fns not present
Neither `unsafe` nor `extern` functions show up in the imenu.

``` rust
fn one() {
    // works
}

unsafe fn two() {
    // doesn't work
}

extern "C" fn three() {
    // doesn't work
}
```

result

mrBliss:
Fixes #188.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -3,6 +3,7 @@
 (require 'rust-mode)
 (require 'ert)
 (require 'cl)
+(require 'imenu)
 
 (setq rust-test-fill-column 32)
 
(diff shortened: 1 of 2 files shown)


mrBliss:
When I have some more time, I plan to refactor this (`rust-re-item-def-imenu`in particular) to be more grammar-like using the `rx` macro. This should make future additions (`static mut FOO: ...` is next on my list) much easier and less messier. Also, currently all modifiers (`pub`, `unsafe`, `extern`) are allowed for all items, so bogus things like `extern trait Foo { ..` are recognized as well. I don't think this is a real problem, but it feels dirty to me.

blaenk:
Thank you!

nikomatsakis:
Looks good :)

PR #191, opened 2017-01-12
fix PR link in README

Wooble:
The markdown was using [] instead of () for the URL, followed by a parenthetical explanation, which was making an ugly relative URL with the explanation text as the link...


pnkfelix:
thanks!

Issue #192, opened 2017-01-17
rust-font-lock-extend-region interferes with jit-lock


DarthFennec:
See [here](https://github.com/DarthFennec/highlight-indent-guides/issues/7) for context.

When `rust-font-lock-extend-region` runs, if the buffer contains a string or comment which contains a `r#*"` (for example, the string `"car"`), it will be interpreted as the beginning of a raw string literal, even though it isn't one. This "raw string" will span until a matching raw string terminator is found, or until the end of the buffer, and the region will be incorrectly extended to that point. This interferes with at least [`highlight-indent-guides`](https://github.com/DarthFennec/highlight-indent-guides), as Emacs never `jit-lock`s these improperly extended regions, and `highlight-indent-guides` relies on `jit-lock` to work properly. I'm sure this causes issues in other modes as well, but I couldn't give you specific examples.

The problem appears to be that `rust-font-lock-extend-region` doesn't check to make sure `(match-beginning 0)` isn't already part of a string or comment. I expect this is simply a matter of adding that check.

PR #197, opened 2017-03-15
Syntax propertize

tromey:


tromey:
I forgot to mention, one other advantage of this approach is that raw strings are fontified as they are typed now, because with the new setup there's no need to match the entire string at once.

DarthFennec:
> I suspect this might fix #192, but I haven't tried it.

I just gave it a try, and it looks like this does fix it. :+1:

nikomatsakis:
@tromey sorry for the long delay here! Let me take a look.

rust-mode-tests.el
```diff
@@ -1377,24 +1372,6 @@ this_is_not_a_string();)"
      "r\" this is a comment\n" font-lock-comment-face
      "\"this is a string\"" font-lock-string-face)))
 
-(ert-deftest font-lock-raw-string-constant ()
-  ;; There was an issue in which a multi-line raw string would be fontified
-  ;; correctly if inserted, but then incorrectly if one of the lines was then
-  ;; edited.  This test replicates how font-lock responds when text in the
-  ;; buffer is modified in order to reproduce it.
```
nikomatsakis: Why remove these tests? I always get nervous when I see that =)

Do they not pass anymore? Or do you just feel they're unneeded now?

rust-mode-tests.el
```diff
@@ -2603,18 +2552,6 @@ Fontification needs to include this whole string or none of it.
     )
   )
 
-(ert-deftest rust-test-revert-hook-preserves-point ()
```
nikomatsakis: Similarly for this test...

nikomatsakis:
@tromey do you mind adding a few more comments as to what's happening here? I'm more of an emacs *user* than developer; I gather you are hooking into the emacs rust-mode machinery in some better way?

@MicahChalmer any chance you can take a look?

rust-mode-tests.el
tromey: This test is actually testing `rust-font-lock-extend-region`, albeit in directly.  That is, the test calls `font-lock-after-change-function`, which then calls that function.  And, this is also testing that the syntax is set correctly, but it's doing so in an indirect way, by looking to see if fontification occurred.

I could try to rewrite the test, but as-is it doesn't work, because the region extension function is gone.  Everything still works fine interactively, though.

rust-mode-tests.el
tromey: This was a test for `rust--after-revert-hook` (I think specifically just for the call to `save-excursion` there), which is gone now.

rust-mode-tests.el
tromey: Oops, I was only talking about the `font-lock-raw-string-constant` test there.

The other test, `font-lock-extend-region-in-string`, explicitly called `rust-font-lock-extend-region`; which is gone, making the test irrelevant, I think.

tromey:
> do you mind adding a few more comments as to what's happening here? I'm more of an emacs user than developer; I gather you are hooking into the emacs rust-mode machinery in some better way?

The basic issue to be solved here involves the fact that Rust has two kinds of angle brackets, and raw strings.

Some angle brackets are paired, and should be treated as pairs (highlight matching, skip over them with sexp commands, etc).  Others are comparison operators and should not be treated this way.

Raw strings have a different delimiter from ordinary strings and in particular it can be multiple characters and symmetric (as opposed to identical).

Emacs has a base feature called a "syntax table" that controls how Emacs sees constructs like matching parens, strings, or comments.  However, syntax tables are limited and can't directly handle the sorts of things that Rust needs.  So, Emacs provides a way to override the syntax of a given character (or group of characters), for cases like this.

In Emacs 23, the way to do this syntax overriding was to make it a side effect of font-lock.  This is what rust-mode was doing.

Starting in Emacs 24, this approach was deprecated, and instead a more direct syntax propertization approach was added.  This works by hooking into `syntax-ppss`, which is the primitive that elisp usually uses to query for the syntax at a given position.  So, this is a bit lower level and doesn't require font-lock.

This patch just changes the syntax propertization approach from the old one to the new one.  Because font-lock uses syntax properties in order to do some face-setting (in particular "syntactic font lock", which is how things like strings and comments are fontified), this new approach won't break font-lock.

The benefits are not dramatic, but it's a bit cleaner and smaller, and seems to fix #192.

MicahChalmer:
I don’t have anything other than agreement to add to @tromey’s explanation above.  (I haven’t had any time to spend on rust, emacs, or anything related to either in a while…). I’m encouraged to see that switching allows the removal of some of the hacks like the revert hook.  The test removals seem appropriate, and the tests that e.g. open a buffer and do a revert and make sure the syntax is detected correctly are still there.

For what it’s worth: when I wrote the code to detect angle brackets, I remember starting off using syntax-propertize, but switching over to font-lock before submitting it, because at the time rust-mode was still supporting emacs 23.

Now that it’s not, I recommend merging this.

nikomatsakis:
OK, looking good, thanks @tromey !

PR #198, opened 2017-03-19
fix rust indentation bug

tromey:
This patch fixes a bug I found where rust-mode would misindent code
like:

fn each_split_within<'a, F>(ss: &'a str, lim: usize, mut it: F)
                            -> bool where F: FnMut(&'a str) -> bool {
}

fn test_split_within() {
}

In particular the second "fn" would be indented a level.

On the "fn" line, rust-mode-indent-line was calling
rust-beginning-of-defun, which would go to the previous defun.  Fixing
this required moving the definition of rust-top-item-beg-re higher in
the file, before its first use (recent versions of Emacs insist on
this).  And, this required removing the "^" from this regexp, which is
ok because the sole use is already adding a "^".

This patch includes the above as a new test case.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -35,6 +35,14 @@
 (defconst rust-re-unsafe "unsafe")
 (defconst rust-re-extern "extern")
 
+;;; Start of a Rust item
+(defvar rust-top-item-beg-re
+  (concat "\\s-*\\(?:priv\\|pub\\)?\\s-*"
+          (regexp-opt
+           '("enum" "struct" "type" "mod" "use" "fn" "static" "impl"
+             "extern" "trait"))
+	  "\\_>"))
+
 (defun rust-looking-back-str (str)
   "Like `looking-back' but for fixed strings rather than regexps (so that it's not so slow)"
   (let ((len (length str)))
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -96,6 +96,13 @@
     )
    ))
 
+;;; Start of a Rust item
+(defvar rust-top-item-beg-re
+  (concat "\\s-*\\(?:priv\\|pub\\)?\\s-*"
```
nikomatsakis: How does this behave with nested items like

```rust
fn foo() {
    fn bar() {
    }
}
```

nikomatsakis: (Can you add a test for that case, if one does not exist already?)

tromey: Yeah, nice catch -- if I wrap the new test case in an outer `fn`, indentation fails once again.

tromey: Ok, I was mistaken about this.  What was happening was that the first draft of my new test case included a blank line, but these aren't currently handled very well by the indentation tests, because those add a six space prefix to each line, which then isn't removed from an otherwise blank line when reindenting.

Still, I've added the new test and made one minor tweak, which is to move the \\_> into the regexp here.

nikomatsakis:
looks like this needs a rebase, but lgtm

tromey:
Merging myself since it is approved.

PR #201, opened 2017-03-22
fix syntax of "<" appearing after "?"

### LINKED ISSUE #200: Wrong indentation with '?' operator and '<' operator
Probably easiest to show an example:

```
fn foo() {
    if bar()? < 1 {
    }
                baz();
}
```
(Also in show-paren-mode, putting point after the final `}` of the function highlights the `<` instead of the opening `{`)

Without the `?`, or with any other operator besides `<`, it indents (and does show-paren) correctly.

tromey:
The "<" syntax-setting code could be confused after a "?".  This patch
changes the code to treat "?" as an "ambiguous operator" and adjust
according to further context.
Fixes #200

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1891,6 +1891,13 @@ fn main() {
 "))
     (test-indent text text)))
 
+(ert-deftest indent-question-mark-operator ()
+  (test-indent "fn foo() {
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Needs rebase =)

tromey:
Done.

PR #206, opened 2017-04-18
Allow formatting with long lines (fixes #186)

### LINKED ISSUE #186: Formatting when `rustfmt` returns an exit code of 3
This happens, for example, when you have a very long URL, string or other long identifier that `rustfmt` can't break up. However, it still formats everything else and puts that in standard output, just as if it had succeeded. Any warnings it prints will go to standard error. Therefore, I believe we should still be able to use `rustfmt` in this case. Currently, the `rust-mode` doesn't work with any `rustfmt` exit code other than 0.

cjhowe7:
It's very annoying to not be able to format files with long comments. This provides a solution to that; but the stderr output is not displayed anywhere. I think that should be added at some point, but it's not really necessary to at least allow formatting these files. The stderr output will let you know which lines exceed the line length, which you might care about if you want to be strict about enforcing line length.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1195,13 +1195,25 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (with-current-buffer (get-buffer-create "*rustfmt*")
     (erase-buffer)
     (insert-buffer-substring buf)
-    (let ((ret (call-process-region (point-min) (point-max) rust-rustfmt-bin t '(t nil) nil)))
-      (if (or (zerop ret) (= ret 3))
-          (progn
-            (if (not (string= (buffer-string) (with-current-buffer buf (buffer-string))))
-                (copy-to-buffer buf (point-min) (point-max)))
-            (kill-buffer))
-        (error "Rustfmt failed, see *rustfmt* buffer for details")))))
+    (let ((tmpf (make-temp-file "rustfmt")))
+      (let ((ret (call-process-region (point-min) (point-max) rust-rustfmt-bin
+                                      t `(t ,tmpf) nil)))
+        (format-message "%d" ret)
+        (cond
+         ((zerop ret)
+          (error "Rustfmt failed, see *rustfmt* buffer for details"))
+         ((= ret 3)
+          (if (not (string= (buffer-string)
+                            (with-current-buffer buf (buffer-string))))
+              (copy-to-buffer buf (point-min) (point-max)))
+          (erase-buffer)
+          (insert-file-contents tmpf)
+          (error "Rustfmt could not format some lines, see *rustfmt* buffer for details"))
+         (t
+          (if (not (string= (buffer-string)
+                            (with-current-buffer buf (buffer-string))))
+              (copy-to-buffer buf (point-min) (point-max)))
+          (kill-buffer)))))))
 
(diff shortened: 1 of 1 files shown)


cjhowe:
To add the stderr output into a buffer, we'd need to create a temporary file, and then load the contents of that temporary file into a buffer.

cjhowe:
I added the stderr output to the `*rustfmt*` buffer. This allows you to see which lines are too long if you wish to fix them.

rust-mode.el
```diff
@@ -1195,12 +1195,24 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
-    (if (zerop (call-process-region (point-min) (point-max) rust-rustfmt-bin t t nil))
-        (progn
-          (if (not (string= (buffer-string) (with-current-buffer buf (buffer-string))))
+    (let ((tmpf (make-temp-file "rustfmt")))
+      (let ((ret (call-process-region (point-min) (point-max) rust-rustfmt-bin
+                                      t `(t ,tmpf) nil)))
+        (cond
+         ((zerop ret)
+          (error "Rustfmt failed, see *rustfmt* buffer for details"))
+         ((= ret 3)
```
nikomatsakis: rustfmt returns *3* in this case? weird. Is this specified somewhere, or it just happens to be true?

cjhowe: Yes, it's documented in the last paragraph here in the README: https://github.com/rust-lang-nursery/rustfmt/blob/master/README.md#running

cjhowe:
I've been using this for a while and I found a bug with some files not formatting correctly. Please don't merge this until I fix it, thanks!

cjhowe:
Of course it was a stupid mistake. Fixed!

cjhowe7:
I've been used this quite a lot now. I'm pretty sure it's ready to be merged.

rust-mode.el
```diff
@@ -1195,12 +1195,24 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (with-current-buffer (get-buffer-create "*rustfmt*")
     (erase-buffer)
     (insert-buffer-substring buf)
-    (if (zerop (call-process-region (point-min) (point-max) rust-rustfmt-bin t t nil))
-        (progn
-          (if (not (string= (buffer-string) (with-current-buffer buf (buffer-string))))
+    (let ((tmpf (make-temp-file "rustfmt")))
+      (let ((ret (call-process-region (point-min) (point-max) rust-rustfmt-bin
```
tromey: It's more idiomatic to use a single `let*` in this case, like
```
(let* ((tmpf ...)
       (ret ...)
```

rust-mode.el
```diff
@@ -1195,12 +1195,24 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (with-current-buffer (get-buffer-create "*rustfmt*")
     (erase-buffer)
     (insert-buffer-substring buf)
-    (if (zerop (call-process-region (point-min) (point-max) rust-rustfmt-bin t t nil))
-        (progn
-          (if (not (string= (buffer-string) (with-current-buffer buf (buffer-string))))
+    (let ((tmpf (make-temp-file "rustfmt")))
```
tromey: It seems like it would be better to wrap the the body here in an `unwind-protect` that removes the temporary file.

tromey:
I had a couple of minor nits but I think it looks good overall.  Thanks for doing this.

cjhowe7:
I made the changes and tested it, and it seems to work fine. Let me know if there's anything else, and thanks for the review!

PR #207, opened 2017-04-22
Adding "default" as keyword

sunchao:


tromey:
I think the patch looks fine, but is `default` unconditionally a keyword?  I skimmed the RFC but it seemed somewhat unclear as to whether it's always a keyword or whether it's only one contextually.  If it's contextual (and TBH I don't even know why I think it could possibly be) then I think some other approach would be preferable.  But if not then this is good.

nikomatsakis:
It is contextual.

sunchao:
Thanks guys. Seems other approach is preferable for this issue. I'll close as invalid now.

Issue #193, opened 2017-01-28
How can I invoke cargo build when I invoke compile in emacs ?


shivanth:
When I invoke compile in rust-mode, emacs runs make command. Is there any configuration to be done to invoke cargo instead of make ?

tromey:
All I do is `M-x compile`, then edit the command line to read `cargo build`.  In my .emacs I have

```
(setq savehist-additional-variables '(compile-command))
(savehist-mode 1)
```

which arranges to save the most recent compile command (the compile command history is automatically saved by savehist).

I haven't found a need for more than this.  Is there something additional that you need?

RalfJung:
I now have emacs do `(setq compile-command "cargo build")` when entering Rust mode. Would it make sense to do something like that per default? `make -k` is pretty much never right for Rust.

Aankhen:
I believe #216 should resolve this.

tromey:
This issue is fixed, so I'm closing it.  But see #226.  I will comment there.

PR #216, opened 2017-06-26
Set `compile-command' in `rust-mode'

Aankhen:
Per @RalfJung in #193.


nikomatsakis:
Seems good to me! Gonna merge. @tromey let me know if you think this is not the right way to do it. =)

tromey:
See #226.

PR #219, opened 2017-07-28
Highlight question mark operator using new `rust-question-mark-face'

### LINKED ISSUE #171: Highlight questionmark
The `?` sigil should be indicated in a very noticeable way. I see two ways to do this:
1. Highlight the sigil itself in a bright color.
2. Put something in the gutter. For example: `?`, `⤴`, `⇡`, or even an image for GUI mode.

IMO a combination of the two would be nice but (1) is obviously much easier.

Aankhen:
This is a pretty small change even with basic tests. I’ve only given the face bold styling and let it inherit from `font-lock-builtin-face`, to avoid making it jarring, but of course users can customize the face.

Related: #171.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1510,6 +1510,38 @@ this_is_not_a_string();)"
    ;; Only the i32 should have been highlighted.
    '("i32" font-lock-type-face)))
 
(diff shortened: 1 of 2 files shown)


tromey:
Looks good, thanks.

Issue #203, opened 2017-04-04
String interpolation different color than strings.


lorenzoi:
String interpolation should be a different color than the rest of the strings. If this is already a feature, what would the syntax for custom themes be?

nikomatsakis:
Are you referring to `format!` etc?

lorenzoi:
no. like 

`println!("{}", x)`

the `{}` and what ever is inside it should be a different color

tromey:
I think this is done.

PR #220, opened 2017-07-28
Highlight interpolation in arguments to print! &c.

Aankhen:
Mark `eprint!`, `eprintln!`, `format!`, `print!`, and `println!` as builtins and highlight interpolation (using italics by default) in the format strings per #203. I’ve also included a special case for `write!`, but I’m uncertain as to whether that’s the correct thing to do since it has to be specifically imported, unlike the other macros. This handles escaped opening braces (`{{`) correctly, but I’m not sure how to deal with escaped closing braces (`}}`).

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -2024,6 +2024,226 @@ fn main() {
    '("r#\"\"\"#" font-lock-string-face
      "'q'" font-lock-string-face)))
 
+(ert-deftest rust-macro-font-lock ()
+  (rust-test-font-lock
+   "foo!\(\);"
+   '("foo!" font-lock-preprocessor-face))
+  (rust-test-font-lock
+   "foo!{};"
+   '("foo!" font-lock-preprocessor-face))
+  (rust-test-font-lock
+   "foo![];"
+   '("foo!" font-lock-preprocessor-face)))
+
+(ert-deftest rust-string-interpolation-matcher-works ()
+  (dolist (test '(("print!\(\"\"\)" 9 11 nil)
+                  ("print!\(\"abcd\"\)" 9 15 nil)
+                  ("print!\(\"abcd {{}}\"\);" 9 19 nil)
+                  ("print!\(\"abcd {{\"\);" 9 18 nil)
+                  ("print!\(\"abcd {}\"\);" 9 18 ((14 16)))
+                  ("print!\(\"abcd {{{}\"\);" 9 20 ((16 18)))
+                  ("print!\(\"abcd {}{{\"\);" 9 20 ((14 16)))
+                  ("print!\(\"abcd {} {{\"\);" 9 21 ((14 16)))
+                  ("print!\(\"abcd {}}}\"\);" 9 20 ((14 16)))
+                  ("print!\(\"abcd {{{}}}\"\);" 9 20 ((16 18)))
+                  ("print!\(\"abcd {0}\"\);" 9 18 ((14 17)))
+                  ("print!\(\"abcd {0} efgh\"\);" 9 23 ((14 17)))
+                  ("print!\(\"{1} abcd {0} efgh\"\);" 9 27 ((9 12) (18 21)))
+                  ("print!\(\"{{{1} abcd }} {0}}} {{efgh}}\"\);" 9 33 ((11 14) (23 26)))))
+    (destructuring-bind (text cursor limit matches) test
+      (with-temp-buffer
+        ;; make sure we have a clean slate
+        (save-match-data
+          (set-match-data nil)
+          (insert text)
+          (goto-char cursor)
+          (if (null matches)
+              (should (equal (progn
+                               (rust-string-interpolation-matcher limit)
+                               (match-data))
+                             nil))
+            (dolist (pair matches)
+              (rust-string-interpolation-matcher limit)
+              (should (equal (match-beginning 0) (car pair)))
+              (should (equal (match-end 0) (cadr pair))))))))))
+
+(ert-deftest rust-formatting-macro-font-lock ()
+  ;; test that the block delimiters aren't highlighted and the comment
+  ;; is ignored
+  (rust-test-font-lock
+   "print!(\"\"); { /* print!(\"\"); */ }"
+   '("print!" rust-builtin-formatting-macro-face
+     "\"\"" font-lock-string-face
+     "/* " font-lock-comment-delimiter-face
+     "print!(\"\"); */" font-lock-comment-face))
+  ;; other delimiters
+  (rust-test-font-lock
+   "print!{\"\"}; { /* no-op */ }"
+   '("print!" rust-builtin-formatting-macro-face
+     "\"\"" font-lock-string-face
+     "/* " font-lock-comment-delimiter-face
+     "no-op */" font-lock-comment-face))
+  ;; other delimiters
+  (rust-test-font-lock
+   "print![\"\"]; { /* no-op */ }"
+   '("print!" rust-builtin-formatting-macro-face
+     "\"\"" font-lock-string-face
+     "/* " font-lock-comment-delimiter-face
+     "no-op */" font-lock-comment-face))
+  ;; no interpolation
+  (rust-test-font-lock
+   "print!(\"abcd\"); { /* no-op */ }"
+   '("print!" rust-builtin-formatting-macro-face
+     "\"abcd\"" font-lock-string-face
+     "/* " font-lock-comment-delimiter-face
+     "no-op */" font-lock-comment-face))
+  ;; only interpolation
+  (rust-test-font-lock
+   "print!(\"{}\"); { /* no-op */ }"
+   '("print!" rust-builtin-formatting-macro-face
+     "\"" font-lock-string-face
+     "{}" rust-string-interpolation-face
+     "\"" font-lock-string-face
+     "/* " font-lock-comment-delimiter-face
+     "no-op */" font-lock-comment-face))
+  ;; text + interpolation
+  (rust-test-font-lock
+   "print!(\"abcd {}\", foo); { /* no-op */ }"
+   '("print!" rust-builtin-formatting-macro-face
+     "\"abcd " font-lock-string-face
+     "{}" rust-string-interpolation-face
+     "\"" font-lock-string-face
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -170,7 +170,18 @@ function or trait.  When nil, where will be aligned with fn or trait."
   "Face for the question mark operator."
   :group 'rust-mode)
 
+(defface rust-builtin-formatting-macro-face
+  '((t :inherit font-lock-builtin-face))
+  "Face for builtin formatting macros \(print! &c.\)."
+  :group 'rust-mode)
+
+(defface rust-string-interpolation-face
+  '((t :slant italic :inherit font-lock-string-face))
```
Wilfred: I find this a little too subtle. groovy-mode and coffee-mode both use colour to highlight interpolations.

rust-mode.el
```diff
@@ -170,7 +170,18 @@ function or trait.  When nil, where will be aligned with fn or trait."
   "Face for the question mark operator."
   :group 'rust-mode)
 
+(defface rust-builtin-formatting-macro-face
+  '((t :inherit font-lock-builtin-face))
+  "Face for builtin formatting macros \(print! &c.\)."
```
Wilfred: You don't need to escape parens in strings unless they're on the start of a line:

```emacs-lisp
"foo (bar)
\(baz) (biz)
```

rust-mode.el
```diff
@@ -573,6 +584,27 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
...
 	     (throw 'rust-path-font-lock-matcher match))))))))
 
+(defvar rust-string-interpolation-re
+  "\\({[^}]*}\\)"
+  "Regular expression to match interpolation syntax in strings.")
+
+(defun rust-string-interpolation-matcher (limit)
+  (unless (rust-in-cmnt)
+    (save-match-data
+      ;; we presently ignore escaping closing delimiters
```
Wilfred: Handling closing delimiters, especially with nesting, is tricky. If you want to handle this, there's an example in groovy-mode: https://github.com/Groovy-Emacs-Modes/groovy-emacs-modes/blob/6713a7d29ba59100db4c168fc6e6bbef277b64ab/groovy-mode.el#L286-L311

Wilfred:
Nice! I'd definitely like to see this.

rust-mode.el
shivjm: I figure it’s better to keep the default styling conservative and let people customize it if they need something more dramatic. That said, the italics have been working really well for me!

Aankhen: I figure it’s better to keep it conservative by default and let users customize it if they want something dramatic. Having said that, the italics stand out nicely for me. I guess it’s different from font to font.

rust-mode.el
Aankhen: Ah, I didn’t know that. I got into the habit of escaping them a long time ago to avoid issues with matching. Thanks!

rust-mode.el
Aankhen: Thanks, those gave me some useful ideas. Is nesting something that Rust implements? I thought it was only escaping.

Aankhen:
Thanks for the feedback, @Wilfred. I’ve removed the spurious escaping you pointed out and added support for escaped closing delimiters in Rust strings—my understanding is that nesting is irrelevant in format strings, so I used simple searching to find them.

rust-mode-tests.el
```diff
@@ -2024,6 +2024,172 @@ fn main() {
...
+        (insert text)
+        (goto-char cursor)
+        (dotimes (n (length matches))
+          (rust-string-interpolation-matcher limit)
+          ;; skip match 0
+          (should (equal (match-string 0) (elt matches n))))))))
+
+(ert-deftest rust-formatting-macro-font-lock ()
+  ; test that the block delimiters aren't highlighted and the comment
+  ; is ignored
```
tromey: In elisp, comments that indent with the code start with `;;`.  A single semi is reserved for a comment to the right of code.

tromey: BTW this is described in `(info "(elisp) Comment Tips")`

rust-mode.el
```diff
@@ -573,6 +584,46 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
             ((not (looking-at (rx (0+ space) "<")))
 	     (throw 'rust-path-font-lock-matcher match))))))))
 
+(defun rust-next-string-interpolation (limit)
+  "Search forward from point for next Rust interpolation marker
+before LIMIT.
+Set point to the end of the occurrence found, and return match beginning
+and end."
+  (while (search-forward "{{" limit t)
```
tromey: It seems to me that this approach can cause the function to skip over formats that come before escapes.  The test would be something like `"{} {{"`.  I think instead there has to be just one loop that looks at the following character to decide what to do.

rust-mode.el
```diff
@@ -573,6 +584,46 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
...
+          (when (search-forward "}" limit t)
+            (throw 'match (list start (point)))))))))
+
+(defun rust-string-interpolation-matcher (limit)
+  "Match next Rust interpolation marker before LIMIT and set
+match data if found. Returns nil if not within a Rust string."
+  (when (rust-in-str)
+    (let ((match (rust-next-string-interpolation limit))) ; too bad `when-let' is 25+
+      (when match
+        (set-match-data match)
```
tromey: I was surprised to see `set-match-data` here; the docs say that it should be called with a list returned by `match-data` (but also confusingly document what `match-data` returns, making this use seem maybe ok).

But looking more it seems to me that the loop in `rust-next-string-interpolation` could be replaced by a single call to `re-search-forward`, and this could set the match data as well..?

rust-mode.el
```diff
@@ -590,6 +641,28 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
      (,(rust-re-grab (concat "#\\!?\\[" rust-re-ident "[^]]*\\]"))
       1 font-lock-preprocessor-face keep)
 
+     ;; Builtin formatting macros
+     (,(concat (rust-re-grab (concat (regexp-opt rust-builtin-formatting-macros) "!")) "[[:space:]]*[({[][[:space:]]*\"")
```
tromey: It seems to me that this doesn't handle raw strings, is that right?

rust-mode-tests.el
Aankhen: Thanks, that node is useful.

rust-mode.el
Aankhen: That’s right. Today I learned that you can use those here, heh. I’ll update it accordingly.

rust-mode.el
Aankhen: Yeah, I was surprised to see the pattern in [groovy-mode](https://github.com/Groovy-Emacs-Modes/groovy-emacs-modes/blob/6713a7d29ba59100db4c168fc6e6bbef277b64ab/groovy-mode.el#L286-L311) too, but it seems to work.

I’m not sure how to use `re-search-forward` to also skip over escapes. I’m trying a different approach to handle the various cases now.

Aankhen:
Thanks for the review, @tromey. I’ve fixed the comment styles, added more tests (including one for `{} {{` and one stress test), fixed the tests to check that we have the exact number of matches we should, improved the handling of escapes to pass those tests, and added support for raw strings. I also pulled a few things out into separate variables and functions to make them easier to deal with.

rust-mode.el
tromey: I think you can use a regexp that matches formats but not escapes, like `"{[^{}]*}"`.

rust-mode.el
```diff
@@ -573,6 +584,54 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
...
+(defun rust-next-string-interpolation (limit)
+  "Search forward from point for next Rust interpolation marker
+before LIMIT.
+Set point to the end of the occurrence found, and return match beginning
+and end."
+  (catch 'match
+    (save-match-data
+      (save-excursion
+        (while (search-forward "{" limit t)
+          (if (eql (char-after (point)) ?{)
```
tromey: This seems better but I think a single regexp search would be best, in particular it would avoid the match data stuff (which seems questionable).

rust-mode.el
```diff
@@ -590,6 +649,22 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
      (,(rust-re-grab (concat "#\\!?\\[" rust-re-ident "[^]]*\\]"))
       1 font-lock-preprocessor-face keep)
 
+     ;; Builtin formatting macros
+     (,(concat (rust-re-grab (concat (regexp-opt rust-builtin-formatting-macros) "!")) (concat rust-formatting-macro-opening-re rust-start-of-string-re))
+      (1 'rust-builtin-formatting-macro-face)
+      (rust-string-interpolation-matcher
```
tromey: It seems to me that this must only mark the first `{}` in a format string.  Is that true?

shivjm: No, the anchored matchers are called repeatedly until point moves outside the limit (which in this case is given by `rust-end-of-string`). `rust-string-interpolation-matcher-works`, `rust-formatting-macro-font-lock`, and `rust-write-macro-font-lock` all include tests for this.

Aankhen: No, the anchored matchers are called repeatedly until point moves outside the limit (which in this case is given by `rust-end-of-string`). `rust-string-interpolation-matcher-works`, `rust-formatting-macro-font-lock`, and `rust-write-macro-font-lock` all include tests for this.

rust-mode.el
Aankhen: I would also prefer to use a single regexp search, but I’m not sure how to do it without lookaheads & lookbehinds while also accounting for escapes. :-S

tromey: Ok, I see what you mean there, sorry I didn't understand that before.  It seems like it could still be done using a regexp with alternation, plus an explicit check for `{{` in the loop.

But ... mainly I wanted this to avoid a possibly-invalid call to `set-match-data`; but I looked through the Emacs sources and I found a couple other cases like this, so let's just go ahead.

tromey:
Thanks for persevering.

Aankhen:
Excellent, thanks!

PR #221, opened 2017-07-30
Add syntax highlighting and imenu support for `union`

Wilfred:
`union` is a contextual keyword, so highlight it in the correct
context. Otherwise, treat `union` similarly to `struct`.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -39,12 +39,18 @@
 (defconst rust-re-vis "pub")
 (defconst rust-re-unsafe "unsafe")
 (defconst rust-re-extern "extern")
+(defconst rust-re-union
+  (rx-to-string
+   `(seq
+    (or space line-start)
+    (group symbol-start "union" symbol-end)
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -577,9 +583,12 @@ the desired identifiers), but does not match type annotations \"foo::<\"."
   (append
    `(
      ;; Keywords proper
-     ("\\_<\\(default\\)[[:space:]]+fn\\_>" 1 font-lock-keyword-face)
      (,(regexp-opt rust-mode-keywords 'symbols) . font-lock-keyword-face)
 
+     ;; Contextual keywords
+     ("\\_<\\(default\\)[[:space:]]+fn\\_>" 1 font-lock-keyword-face)
```
tromey: I was under the impression that this rule had to come before the keyword rule.  But if the tests pass, I guess it's ok.

tromey:
Thanks for the patch.

PR #225, opened 2017-08-03
Use `font-lock-variable-name-face' for `let' bindings

Aankhen:
This is an updated version of #173 that includes the changes requested and further expands the tests. The original PR hasn’t been updated since January.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1284,12 +1284,46 @@ list of substrings of `STR' each followed by its face."
    '("fn" font-lock-keyword-face
      "foo_Bar" font-lock-function-name-face)))
 
+(ert-deftest font-lock-let-bindings ()
+  (rust-test-font-lock
+   "let foo;"
+   '("let" font-lock-keyword-face
+     "foo" font-lock-variable-name-face))
+  (rust-test-font-lock
+   "let mut foo;"
+   '("let" font-lock-keyword-face
+     "mut" font-lock-keyword-face
+     "foo" font-lock-variable-name-face))
+  (rust-test-font-lock
+   "let foo = 1;"
+   '("let" font-lock-keyword-face
+     "foo" font-lock-variable-name-face))
+  (rust-test-font-lock
+   "let mut foo = 1;"
+   '("let" font-lock-keyword-face
+     "mut" font-lock-keyword-face
+     "foo" font-lock-variable-name-face))
+  (rust-test-font-lock
+   "fn foo() { let bar = 1; }"
+   '("fn" font-lock-keyword-face
+     "foo" font-lock-function-name-face
+     "let" font-lock-keyword-face
+     "bar" font-lock-variable-name-face))
+  (rust-test-font-lock
+   "fn foo() { let mut bar = 1; }"
+   '("fn" font-lock-keyword-face
+     "foo" font-lock-function-name-face
+     "let" font-lock-keyword-face
+     "mut" font-lock-keyword-face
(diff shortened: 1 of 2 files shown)


nordlow:
Thanks for doing this.

rust-mode-tests.el
```diff
@@ -1555,8 +1594,9 @@ this_is_not_a_string();)"
   (rust-test-font-lock
    "let union = 7; union foo { x: &'union bar }"
    '("let" font-lock-keyword-face
-     ;; Ignore the first union, it's an unhighlighted variable.
+     ;; Ignore the first union, it's a variable.
```
tromey: I think the comment should change, because the first occurrence of union is no longer being ignored.

tromey:
Thanks.  This looks good.  I had just one nit.

Also I was wondering if it is possible to preserve the original commits or mention the original author's name in the commit message.  I think it's nice to preserve authorship information when possible; at the very least to give credit, but also occasionally there are other reasons.

Hi-Angel:
> Also I was wondering if it is possible to preserve the original commits or mention the original author's name in the commit message.

It's possible: `git am nordflow-commit.patch` preserves the authorship of the patch, and then further commits can be made as usual.

rust-mode-tests.el
Aankhen: Gotcha.

Aankhen:
@nordlow:

> Thanks for doing this.

No sweat.

@tromey:

> Also I was wondering if it is possible to preserve the original commits or mention the original author's name in the commit message. I think it's nice to preserve authorship information when possible; at the very least to give credit, but also occasionally there are other reasons.

Yes, good point. I should have done that myself.

@Hi-Angel:

> It's possible: git am nordflow-commit.patch preserves the authorship of the patch, and then further commits can be made as usual.

Thanks! Is this different from merging nordlow’s branch before my commit? `git am` fails because the patch won’t apply against master, whereas I was able to merge and add my commit (plus the merge commit), which gives me nordlow’s commit with its orginal metadata (including the date).

Hi-Angel:
It's not very different from merging their branch first, and then making your commit. Either way in the history of commits would be their commit first, and then your commit. May be there are other ways too — I dunno, I just quickly typed it out off my head so that nobody would search for how to do it.

> `git am` fails because the patch won’t apply against master

I remember me neither could apply it, and when I went to see the reason, turned out it's mainly whitespace changes, the functional change is very small. So given you have the problem, I think you can simply remove from the patch everything except the functional changes. And then may be make them part of your commit; or not; whatever you prefer. I'm sure @nordlow won't get offended because the trailing whitespace removal is for sure automatic, e.g. `(delete-trailing-whitespace)` in Emacs.

Hi-Angel:
Ah, and note: if you have the `(delete-trailing-whitespace)` on your `before-save-hook`, then you better off disable it before modifying the patch, because the patch format is using whitespace for marking some fields. I know because I have already stumbled upon it :) I ended up simply modifying the patch with vim from a terminal.

Aankhen:
It seems like the difference is that a non–fast-forward `git merge` will add a merge commit and put the actual commits in their original places in the log, whereas `git am` will directly append the commits to the log. After a lot of experimentation, I got it working by using `git rebase`, going back to after the last commit on rust-lang:master, using `git am` to pull in the changes (with an edited patch), then continuing the rebase to pull in my changes. Phew!

@tromey, does this look okay now?

tromey:
Thank you.

PR #230, opened 2017-08-14
Fix recognition of "<" as operator in some context

### LINKED ISSUE #212: Incorect Indentation
Rust-mode cannot format this code correctly. Rustfmt works well.

```rust
fn main() {
    let a = [[0u32, 0u32]; 1];
    let i = 0;
    let x = a[i][(1 << i)]; // bad
    }
```

tromey:
rust-mode identifies the "<<" as open angle brackets in
    let x = a[i][(1 << i)];
This patch fixes the problem by changing rust-is-in-expression-context
to treat "]" as starting an expression in brace context.

Fixes #212

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -2034,6 +2034,17 @@ fn main() {
     baz();
 }"))
 
+;; Regression test for #212.
+(ert-deftest indent-left-shift ()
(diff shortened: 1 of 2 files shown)


tromey:
Merging this per the discussion in the bug.

PR #233, opened 2017-08-20
Add support for u128 and i128 types

shanavas786:


tromey:
Looks good, thanks.  I somewhat prefer clean patches -- no extraneous changes.  But considering that it's just some whitespace...

PR #234, opened 2017-08-20
Types get correct font-lock in if-let statements

### LINKED ISSUE #232: Types using variable-font-lock-face in if-let bindings
I noticed that after updating rust-mode types in if-let bindings now uses the variable font lock instead of the proper type font lock. Is this intended behavior? If so, for what reason? Cheers!

jjwest:
Changed so that font-locking for types is considered before font-locking for variable names in let bindings.
Fixes #232

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -2345,6 +2345,15 @@ fn main() {
      "/* " font-lock-comment-delimiter-face
      "no-op */" font-lock-comment-face)))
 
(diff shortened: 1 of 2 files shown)


tromey:
It seems fine to me, but would you mind adding a regression test for this?

jjwest:
Added a test.

tromey:
Thank you for the patch.

PR #235, opened 2017-08-23
Pass 'visible to get-buffer-window

tromey:
If I have two frames, with the *compilation* buffer in one frame and
use next-error in the other frame, I get this error:

Wrong type argument: window-live-p, nil

This happens because get-buffer-window returns nil.  The fix is to
pass 'visible to get-buffer-window, which will pick any visible frame
from the same terminal.


tromey:
This one is pretty trivial and I've been using it for a while, so I'm just going to merge it.

Issue #204, opened 2017-04-05
rustfmt should display errors


nikomatsakis:
If you run rustfmt and it errors out, it currently just "beeps" and suggests you switch over to `*rustfmt*`. I think this should be treated like a failed compilation; we should split the frame into two windows, show the errors, and let you use `M-x next-error` to view them.

Example 1, which errors because it doesn't parse:

```rust
fn main() {
    let x &'static str = "aaa";
}
```

Example 2, undefined module:

```rust
mod a;

fn main() {
    let x: &'static str = "aaa";
}
```

Example 3, long lines:

```rust
fn main() {
    let x:
    &'static str = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
}
```

brotzeit:
I've tried to solve this with `shell-command-on-region`, but it doesn't work for example 2 as the function returns 0 in this case. Does anybody know why ?


```emacs-lisp
;; Formatting using rustfmt
(defun rust--format-call (buf)
  "Format BUF using rustfmt."
(let ((fmt-buffer (get-buffer-create "*rustfmt*")))
    (with-current-buffer fmt-buffer
      (let ((buffer-read-only nil))
        (erase-buffer)
        (insert-buffer-substring buf)
        (let ((ret (shell-command-on-region (point-min) (point-max)
                                            rust-rustfmt-bin fmt-buffer
                                            nil)))
          (cond
           ((zerop ret)
            (if (not (string= (buffer-string)
                              (with-current-buffer buf (buffer-string))))
                (copy-to-buffer buf (point-min) (point-max)))
            (kill-buffer)
            (message "Formatted buffer with rustfmt."))
           (t
            (let ((path (buffer-file-name buf)))
              (goto-char (point-min))
              (while (re-search-forward "stdin" nil t)
                (replace-match path))
              (compilation-mode)
              (goto-char (point-min))
              (pop-to-buffer fmt-buffer)
              (message "Rustfmt could not format some lines.")))))))))

```

tromey:
> Does anybody know why ?

All I really know is that `shell-command-on-region` returns the exist code of the subprocess.  So it must be doing `exit 0`.

[...]

PR #236, opened 2017-08-25
display rustfmt errors in compilation buffer

brotzeit:
As I mentioned in  #204, I can't display errors regarding undefined modules. I found the reason in [#562](https://github.com/rust-lang-nursery/rustfmt/issues/562) and I don't see a possibility to fix this without saving the file or creating a temp file. Any ideas ?


tromey:
> I don't see a possibility to fix this without saving the file or creating a temp file. Any ideas ?

Maybe a temporary file isn't too bad.

Issue #226, opened 2017-08-04
compile-command=`cargo build` breaks workflow


Dushistov:
After #216 my workflow become broken.

For example,

1. I open lib.rs
2. `M-x compile` and choose `cargo check`
3. jump to buffer with another file in the same crate and invoke `M-x recompile`,
   `emacs` instead of `cargo check` calls `cargo build`,

This is especially wrong for crates that may be compiled only with specific `target` and not default one. So if type `cargo build   --target=arm-linux-androideabi` and then recompile
you got compile time error, because it invokes `cargo build` instead of `cargo build   --target=arm-linux-androideabi`.

tromey:
I tend to think that that patch should be reverted.  People who really want to force `compile-command` like that can easily do it in their mode hook.  But, it seems clear now that the change is not appropriate for all cases.

RalfJung:
So having to customize the mode hook is expected behavior? :/

I suppose that's just how emacs works, and this mode cannot do much about that. In my experience, "customize the mode hook" is equivalent to "random mutation of lisp code until it does what I want it to do". I had hoped this one particular customization maybe could be avoided. Oh well. (These problems are the reason why I now avoid using emacs if I can...)

tromey:
> So having to customize the mode hook is expected behavior? :/

Yeah, it's pretty normal.  I think the problem with the approach taken here is that it sets `compile-command` locally, so that any changes one makes don't apply across the other buffers in the project.  That seems like a minus.

One alternate approach that might work would be to push the cargo build command onto `compile-history`, maybe with some heuristic like "topmost entry doesn't already start with `"cargo"`".  This would make `M-x compile` default to `cargo build` but without interfering with any changes that have been made.

Aankhen:
Oops. Sorry about that. Let me try @tromey’s alternative.

tromey:
That would be reasonable as well, I think.  Nothing even says we have to use compilation mode; it is convenient but there are other bugs showing how it doesn't always line up exactly with what rust does.

iwillspeak:
A temporary work around for this:

```elisp
(add-hook 'rust-mode-hook
          (lambda ()
            (kill-local-variable 'compile-command))
```

rightfold:
It should be possible to set a flag that prevents rust-mode from ever touching `compile-command`. I set `compile-command` myself on each Emacs session depending on what project I am working on (typically `(cd ~/... && cargo test)`), and bind `<f5>` to `compile`.

[...]

PR #237, opened 2017-09-07
Make Compile Command Configurable

iwillspeak:
This allows users to disable the compile command if it inteferes with
their workflow.

Part of #226

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -160,6 +160,12 @@ function or trait.  When nil, where will be aligned with fn or trait."
   :type 'string
   :group 'rust-mode)
 
+(defcustom rust-compile-command "cargo build"
+  "Default compile command when editing Rust."
+  :type '(choice (const :tag "Disabled" nil)
+				 (string :tag "Command"))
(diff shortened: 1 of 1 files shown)


tromey:
The patch is reasonable but I would like to know what you think about just removing that code; or, alternatively, changing the `setq-local` in `rust-mode` to something like (untested):
```
(unless (member rust-compile-command compile-history)
  (push rust-compile-command compile-history))
```
I tend to think that per-project compilation settings should be done with some other facility and that this code should just be removed.

iwillspeak:
If removing it is an option then I vote for that. I just made this pull request as a quick way so that I could disable it but the default behaviour wouldn't change from what it is now.

I am tempted to suggest not trying to set the compile command with some kind of heuristic. Heuristics can be wrong and when they are it just ends up being a right pain for the person using it. Normally you only have to set the compile command once, when you first compile. Then it will be in the history and you can use `M-x recompile`.

iwillspeak:
Closing this in favour of #243

PR #239, opened 2017-09-14
Remove `float`, `int`, and `uint` as built-in types.

shepmaster:
These went away pre Rust-1.0 and are valid identifiers. Went ahead and
removed them from the tests as well, replacing them with real types of
the same number of characters.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -316,11 +316,11 @@ very very very long string
 (ert-deftest indent-struct-fields-aligned ()
   (test-indent
    "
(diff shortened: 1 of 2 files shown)


shepmaster:
For reference, I was looking at some code like

```rust
pub use num::bigint::BigUint;

impl From<BigUint> for Literal {
    fn from(uint: BigUint) -> Self { Literal::Integer(uint) }
}
```

As was confused that the argument name and type were highlighted the same :-)

tromey:
Thank you for the patch.

PR #240, opened 2017-09-21
feat(rust-mode): add yield to the keywords list

michael-zapata:
Enables syntax highlighting for that token :)


tromey:
The patch itself is fine; I'm not following language evolution super-closely, so can you point me at where this is defined?

michael-zapata:
Sure !

It was introduced in rust-lang/rust#43076.

tromey:
Thank you very much.

PR #243, opened 2017-10-03
Revert #216 and Add Suggestions from #226

iwillspeak:
This reverts the `setq-local` for `compile-command` and replaces it
with a function which can be used to invoke compile with `"cargo
build"` as the compile command.


nikomatsakis:
I think we should do this, yes.

PR #244, opened 2017-11-06
% as punctuation character.

kubo39:


tromey:
Thank you.  Funny how that's been missed all this time.

PR #245, opened 2017-11-06
Use evm and test emacs 24.1~25.3/snapshot and remacs-snapshot.

kubo39:
Test passed in https://travis-ci.org/kubo39/rust-mode/builds/297950846

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,4 @@ script:
   - EMACS=`which emacs` ./run_rust_emacs_tests.sh
 
 notifications:
-  email:
-    on_success: never
+  email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,32 @@
 language: generic
 
-env:
-  matrix:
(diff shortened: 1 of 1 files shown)


tromey:
Could you explain what the main benefit of this change would be?  That would be helpful.

.travis.yml
```diff
@@ -1,27 +1,31 @@
 language: generic
 
-env:
-  matrix:
-    - EMACS=emacs24
-    - EMACS=emacs-snapshot
+before_install:
+  - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh
```
pnkfelix: My spidey-sense says that its dangerous to pull from users gist and run it a a shell script.

pnkfelix:
I cannot tell what the benefit of this is, and I am skeptical about the change itself (e.g. I already noted the questionable running of a user gist as a shell script.)

Closing.

PR #246, opened 2017-11-06
use compilation buffer for rustfmt errors

brotzeit:


tromey:
I like this idea, but I wonder about removing the "exit code 3" handling.  IIRC that is there because sometimes rustfmt can't format specific lines -- so currently the changes are applied and a warning is given, but with this change, the changes won't be applied.

brotzeit:
I forgot about that problem. That was actually the reason I closed a previous PR.
The updated solution is not very pretty but I honestly didn't see any other error with exit code 3 than "line exceeded maximum width". 

I will use this solution, but I have no problem with changing it again if anybody has a better idea. But IMO this is much more convenient.

rust-mode.el
```diff
@@ -1332,29 +1332,49 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 ;; Formatting using rustfmt
 (defun rust--format-call (buf)
   "Format BUF using rustfmt."
-  (with-current-buffer (get-buffer-create "*rustfmt*")
-    (erase-buffer)
-    (insert-buffer-substring buf)
-    (let* ((tmpf (make-temp-file "rustfmt"))
```
tromey: Do you know why this code previously used a temporary file? I wonder if there was some issue with using stdin.

rust-mode.el
```diff
@@ -1332,29 +1332,49 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
+                  (line-error "error: line exceeded maximum width[[:ascii:]]+")
+                  buf-string)
+              (copy-to-buffer buf (point-min) (point-max))
+              (with-current-buffer buf
+                (while (re-search-forward line-error nil t)
+                  (replace-match ""))
+                (setq buf-string (buffer-string)))
+              (goto-char (point-min))
+              (save-excursion
+                (while (re-search-forward buf-string nil t)
```
tromey: I don't understand what this loop is for.  Could you add a comment explaining it?

rust-mode.el
brotzeit: Maybe this rust-lang-nursery/rustfmt#562 ?

rust-mode.el
brotzeit: This removes the actual file contents from the compilation buffer, but you're right that can be done better. IMO rustfmt shouldn't format files if there are any kind of errors. I mean I have to take care of them anyway as I don't want an error message every time I use rustfmt. Hopefully that will be changed when it's more mature.

I can understand if you don't want to merge it, as it won't work correctly in every situation. I can try to make this more sound, but I can also wait until there's an improvement on the rustfmt side and give it another try.

nikomatsakis: I know nothing, but I'm curious why we can't use [`compilation-start`](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/compile.el#L1633)?

rust-mode.el
brotzeit: Ok this can't be it, as the old version also uses stdin. Maybe the author preferred `call-process-region` or didn't know about `shell-command-on-region`.

rust-mode.el
brotzeit: I think `compilation-start` doesn't allow piping through stdin.

PR #249, opened 2017-11-18
fix rust--format-call

### LINKED ISSUE #248: *rustfmt* buffer remains empty after rustfmt failure
I'm unable to get anything to show in the rustfmt buffer when I run rustmft against a file that makes rustfmt fail; it's the same whether I do it manually or trigger with a save. I made a breaking edit to an example program:
image
when I save, I get a message that rustfmt failed, as expected:
image
then when I open the buffer-menu to go see what the formatting problem is, \*rustfmt\* is only ever empty:
image
when really I'm expecting it to have this info:
image
are my expectations of what should be in the rustfmt buffer after a formatting failure incorrect? Right now I am disabling rustfmt-on-save, saving the file, opening a shell, then running rustfmt against the file from the shell to get info on where my formatting is messed up. This works but is quite tedious.

brotzeit:
closes #248

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1353,6 +1353,8 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
             (insert-file-contents tmpf)
             (error "Rustfmt could not format some lines, see *rustfmt* buffer for details"))
            (t
+            (erase-buffer)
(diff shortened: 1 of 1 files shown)


tromey:
Thanks!

PR #251, opened 2017-11-24
don't set-window-start for selected window

### LINKED ISSUE #231: rust-enable-format-on-save + compile = annoying jump to begin of file
If I 

1. Enable calling of `rustfmt` before save with `M-x rust-enable-format-on-save`
2. Maximize emacs window
3. Edit code in buffer in such way that coding style not match `rustfmt` (so `rustfmt` will change something)
4. Call `M-x compile` or `M-x recompile`

then cursor jump to begin of file.

For example if I edit line 600 of my rust file,
and reformat `))` to `)\n)`, then command to emacs `M-x compile`,
cursor jump to line 8.

This is very annoying and frustrating.
Have no idea why line 8, there is only one similarity - the first warning in compilation window starts on line 8 from begin of `*compilation*`.

I was abble to fix it for me, by hacking `rust-format-buffer` in such way:

```diff
-      (dolist (loc window-loc)
-        (let* ((window (pop loc))
-               (buffer (window-buffer window))
-               (start (rust--format-get-pos buffer (pop loc)))
-               (pos (rust--format-get-pos buffer (pop loc))))
-          (set-window-start window start)
-          (set-window-point window pos)))))
+      ;; (dolist (loc window-loc)
+      ;;   (let* ((window (pop loc))
+      ;;          (buffer (window-buffer window))
+      ;;          (start (rust--format-get-pos buffer (pop loc)))
+      ;;          (pos (rust--format-get-pos buffer (pop loc))))
+      ;;     (set-window-start window start)
+      ;;     (set-window-point window pos)))
+      )
+    )
```

If I disable rust-format on save I can not reproduce error.

/cc @tspiteri

brotzeit:
closes  #231

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1475,7 +1475,8 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
                (buffer (window-buffer window))
                (start (rust--format-get-pos buffer (pop loc)))
                (pos (rust--format-get-pos buffer (pop loc))))
-          (set-window-start window start)
+          (unless (eq buffer current)
+            (set-window-start window start))
           (set-window-point window pos)))))
 
   (message "Formatted buffer with rustfmt."))


rust-mode.el
```diff
@@ -1475,7 +1475,8 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
                (buffer (window-buffer window))
                (start (rust--format-get-pos buffer (pop loc)))
                (pos (rust--format-get-pos buffer (pop loc))))
-          (set-window-start window start)
+          (unless (eq buffer current)
+            (set-window-start window start))
           (set-window-point window pos)))))
```
brotzeit: Is it possible that `(set-window-point window pos)` is redundant as we already set the position for each buffer in the previous loop ?

tromey: I am not sure, but I was curious why wrap the call to `set-window-start` but not the one to `set-window-point`.

brotzeit: I'm also not sure, but I think `set-window-point` restores a saved location ? `set-window-start` seems to be the reason for  #231 so I only wrapped this function.

tromey:
Thanks for the patch.

PR #252, opened 2017-11-24
use after-save-hook for rust-format-on-save

### LINKED ISSUE #166: Cannot save if rust-format-on-save and rustfmt fails
If you have

```
  (setq rust-format-on-save t)
```

and your code has a syntax error in it anywhere, you can no longer save your file.
Instead when you save you get a "rustfmt failed, see buffer for details" message.

And then in the rustfmt buffer it typically just includes the entire text of your buffer, it's not super helpful.

I think it'd be better to always save, and just warn you that rustfmt failed.

brotzeit:
closes #166

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1552,8 +1552,7 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (setq-local parse-sexp-lookup-properties t)
   (setq-local electric-pair-inhibit-predicate 'rust-electric-pair-inhibit-predicate-wrap)
 
-  (add-hook 'before-save-hook 'rust--before-save-hook nil t)
-
+  (add-hook 'after-save-hook 'rust--after-save-hook nil t)
   (setq-local rust-buffer-project nil)
 
   (when rust-always-locate-project-on-open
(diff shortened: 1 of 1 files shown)


mrBliss:
Saving before rustfmt is run is just more work for the user because they have to save the file again after the reformatting. Furthermore, after this second save rustfmt will be run a second time. So you're doubling the number of saves and the times rustfmt is run.

I think a better approach would be to not return an error in the `after-save-hook`, but just report the failure and continue saving.

brotzeit:
Good point, I didn't think of it. Thanks.

rust-mode.el
```diff
@@ -1568,8 +1567,13 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
   (rust-mode))
 
-(defun rust--before-save-hook ()
-  (when rust-format-on-save (rust-format-buffer)))
+(defun rust--after-save-hook ()
+  (when rust-format-on-save
+    (condition-case rustfmt-err
+        (and
+         (rust-format-buffer)
+         (save-buffer))
```
tromey: Calling `save-buffer` from an `after-save-hook` seems fishy to me.  Maybe it would be better to just `message` or `warn` the failure and not propagate the exception.

brotzeit: I think I remember my problem here. I don't know how to save a file without the message in the minibuffer.

brotzeit: Using `after-save-hook` or `before-save-hook` ?

brotzeit:
I think this has to be done in `before-save-hook`. 
I'm using this and it seems to work.
```elisp
(defun rust-before-save-hook ()
  (when rust-format-on-save
    (let ((proc (rust-format-buffer)))
      (while (eq (process-status proc) 'run)
        (sit-for 0.01)))))
```
But I don't know how this can be achieved with `rust-format-buffer` of rust-mode, sorry.

PR #253, opened 2017-12-08
Improved font-locking for print macros

jjwest:
Improved the font-locking for print macros so that they still apply in cases where the body is empty, 
such as `println!()`, and when the content is on a new line.

Before:
https://i.imgur.com/b6HGC3d.png

After:
https://i.imgur.com/TGNXlMN.png


nikomatsakis:
seems good =)

PR #254, opened 2017-12-10
Fix warning message parsing

finalpatch:
This addresses https://github.com/rust-lang/rust-mode/issues/156

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1591,14 +1591,24 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   "Specifications for matching errors in rustc invocations.
 See `compilation-error-regexp-alist' for help on their format.")
 
-(defvar rustc-new-compilation-regexps
-  (let ((file "\\([^\n]+\\)")
+(defvar rustc-new-error-regexps
+  (let ((message "\\(?:\\[[0-9A-Z]+\\]\\)?: .*\n *--> ")
+        (file "\\([^\n]+\\)")
         (start-line "\\([0-9]+\\)")
         (start-col  "\\([0-9]+\\)"))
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -1591,16 +1591,25 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   "Specifications for matching errors in rustc invocations.
 See `compilation-error-regexp-alist' for help on their format.")
 
-(defvar rustc-new-compilation-regexps
-  (let ((file "\\([^\n]+\\)")
+(defconst rustc-new-compilation-regexps-template
+  (let ((message "\\(?:\\[[0-9A-Z]+\\]\\)?: \\(.*\\)\n *--> ")
```
nikomatsakis: hmm so there is a problem -- iirc, multiline regex do not reliably work in compilation-mode. It took @pnkfelix and I some time to track it down, but what we found was that the compilation-mode would buffer line-by-line, and hence it would sometimes try to match the regex when the "error: ..." line was available, but not the `--> line`. This is the whole reason that we have the hook that "scrolls the window up" after you select a `-->` line as well.

pnkfelix: Yes multi line regexps are broken. And worse, they are broken in a way where the failure is transient/nondeterministic.

I have an open bug against compile.el on the matter here:

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25133

pnkfelix: (Note that i do not think the buffering is strictly line-based... i think it is instead working in 500 character chunks... see previously linked emacs bug.)

pnkfelix:
Closing as we do not trust multiline regexp support in emacs

PR #263, opened 2018-02-15
remove old-style compilation regex; support `:::` regex

nikomatsakis:
The "new-style" error messages are pretty well ingrained by this point, I think.

This also adds a new regex that makes it possible to select the `:::`-style notes that arise from errors inside of macro invocations (see screen shot below):

```
error[E0026]: struct `traits::query::normalize::NormalizationResult` does not have a field named `ambiguity`
   --> src/librustc/macros.rs:255:28
    |
214 | |         }
    | |________________________________^ struct `traits::query::normalize::NormalizationResult` does not have field `ambiguity`
...
255 |                   let $s { $($field,)* } = self;
    |  ____________________________^
    |
   ::: src/librustc/traits/query/normalize.rs
    |
212 | / BraceStructTypeFoldableImpl! {
213 | |     impl<'tcx> TypeFoldable<'tcx> for NormalizationResult<'tcx> {
214 | |         normalized_ty, ambiguity
215 | |     }
216 | | }
    | |_- in this macro invocation
```

screen shot 2018-02-15 at 5 04 02 am

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1585,10 +1585,12 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 See `compilation-error-regexp-alist' for help on their format.")
 
 (defvar rustc-colon-compilation-regexps
-  (let ((file "\\([^\n]+\\)"))
-    (let ((re (concat "^ *::: " file ; ::: foo/bar.rs
+  (let ((file "\\([^\n]+\\)")
+        (start-line "\\([0-9]+\\)")
+        (start-col  "\\([0-9]+\\)"))
+    (let ((re (concat "^ *::: " file ":" start-line ":" start-col ; ::: foo/bar.rs
                       )))
-      (cons re '(1))))
+      (cons re '(1 2 3))))
   "Specifications for matching `:::` hints in rustc invocations.
 See `compilation-error-regexp-alist' for help on their format.")
 
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1574,31 +1574,22 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 (defun rust--before-save-hook ()
   (when rust-format-on-save (rust-format-buffer)))
 
-;; Issue #6887: Rather than inheriting the 'gnu compilation error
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Of course, it'd be nicer if you could jump to the right line number, but I'm not smart enough to do that.

brotzeit:
Some thoughts on this.

I've tried `next-error` without `rustc-scroll-down-after-next-error` and it seems to work.

A path preceded by `:::` is the location of the definition right ? I'd they this shouldn't be treated as an error 
by `next-error` and `compilation-next-error`.

I'm sure this could be done with a regexp in elisp but aren't the line and the colon usually provided by the compiler ?

nikomatsakis:
So, it seems like the (nightly) compiler now emits `:::` errors with line-number too, which is awesome, but makes these regex out of date.

nikomatsakis:
> I've tried next-error without rustc-scroll-down-after-next-error and it seems to work.

Not directly related to this PR, I guess, but: it depends on what you mean by work -- iirc, it will jump to the right place, but the `*compilation*` buffer doesn't show you the current error at the top like it ought to (but rather the `--> foo.rs:22` line). What is the top line of the `*compilation*` buffer that you see?

> A path preceded by ::: is the location of the definition right ? I'd say this shouldn't be treated as an error by next-error and compilation-next-error.

I don't understand this...can you give an example?

brotzeit:
> Not directly related to this PR, I guess, but: it depends on what you mean by work -- iirc, it will jump to the right place, but the *compilation* buffer doesn't show you the current error at the top like it ought to (but rather the --> foo.rs:22 line). What is the top line of the *compilation* buffer that you see?

I could have figured that out by myself. Thanks.

> I don't understand this...can you give an example?

I've only seen the :::-style notes when the following location contains the definition of a function.
This is a screenshot of the `rust-compilation-mode` I have added.

image

nikomatsakis:
updated to include filename / line-numbers

nikomatsakis:
Merging with r=tromey (tested locally, and it works)

PR #266, opened 2018-03-23
throw error in after-save-hook instead of before-save-hook

brotzeit:


brotzeit:
I forgot rust-mode also throws an error for a format error. Maybe it should be a message instead of an error.

pnkfelix:
Hi thanks for the PR!

Any chance you could address the Travis warning??

Issue #268, opened 2018-04-29
rust-format-on-save not respecting local rustfmt.toml


ralphtheninja:
I have the following lines in my `~/.emacs`:

```lisp
;; load rust-mode
(add-to-list 'load-path "/home/lms/dotfiles/emacs/site-lisp/rust-mode")
(autoload 'rust-mode "rust-mode" nil t)
(add-to-list 'auto-mode-alist '("\\.rs\\'" . rust-mode))
(setq rust-always-locate-project-on-open t)
(setq rust-indent-offset 2)
(setq rust-format-on-save t)
```

I'm aware that most projects are using 4 spaces but I need to get 2 space indentation working for a particular project.

My local `rustfmt.toml` looks like:

```toml
max_width = 80
tab_spaces = 2
```

When saving a file in emacs the indentation is still 4 spaces. However, when I manually do `cargo fmt` on the command line, it formats the code as expected.

Any ideas?

ninrod:
Cannot reproduce. I've placed a rustfmt.toml in the git project root and all works fine.

ralphtheninja:
It wasn't related to the emacs lisp. Fixed now :)

Issue #256, opened 2017-12-13
rust-enable-format-on-save + compile = annoying jump to end of buffer


Dushistov:
This is bug is similar to #231, but instead of begin of buffer,
at now it jumps to the end of buffer with rust code.
I have applied fix from #251 

Steps to reproduce:

1. Split window by two and run compile command in the second one:
   `C-x 2` `C-x O` `M-x compile` choose and run compilation command
2. Return back to code window and choose code at the middle of buffer and change it's 
    formating in so way that `rustfmt` should modify it (Do not save buffer)
3. Jump to compilation window `C-x O` and press `g` 

as result the position of cursor in buffer with rust code will change to the end of buffer.

/cc @brotzeit

brotzeit:
Can you do me a favor and try this with my [fork](https://github.com/brotzeit/rust-mode-reloaded). I have added a custom process and it's possible that this issue doesn't occur with it. If the problem still exists I will try to fix it.

Dushistov:
@brotzeit 

I tried. For real project is does not work at all.

```
error[E0583]: file not found for module `cpp_code`
 --> /home/evgeniy/projects/rust-infra/swig/macroslib/src/cpp/mod.rs:1:5
  |
1 | mod cpp_code;
  |     ^^^^^^^^
  |
  = help: name the file either cpp_code.rs or cpp_code/mod.rs inside the directory "/tmp"
```

looks like it copy just one file into '/tmp' and run rustfmt for it?

But for dummy test project created with `cargo --new bin` it looks like work.

I wonder how scheme with copying will work for real projects, with tree of modules,
plus part of code is generated by `build.rs` and lay in `target` and included via `include!`?

brotzeit:
Thanks for taking a look at it!
Actually you wouldn't even get this error with the current solution in rust-mode, because it is ignored when running rustfmt with stdin. I've tried it with a temp file as there are several issues with stdin, but I didn't think of this error.

What do you think would be the best solution. Switch to stdin again, try to ignore this error with the current solution or create a temp file in the project directory ?
I don't like the solution with the temp file, but I think stdin is currently not really usable.

Dushistov:
@brotzeit 

>What do you think would be the best solution. Switch to stdin again, try to ignore this >error with the current solution or create a temp file in the project directory ?
>I don't like the solution with the temp file, but I think stdin is currently not really usable.

[...]

PR #269, opened 2018-05-06
Fix window scrolling after rustfmt is run

apiraino:
Hello,

I tried investigating the window scrolling after rustfmt is run (this issue has been referenced in various forms in issues #268 #231 #251) as I am still experiencing it and imo it's pretty annoying.

I'm not an elisp expert, but I think a quick fix/workaround could be to simply nail the window start position and restore it after rustfmt has finished.

I think there may be still cases where the code reformatting can be a bit confusing (f.e. `rustfmt`'ing a big messed up file), but at least I'm not getting mad every time I compulsively save a buffer :-)

Can you let me know what you think of this little patch? I tested it a bit and it has greatly improved my coding experience with rustfmt.

Thanks for your time!

EDIT: I did some checks, but I'm not sure how to fix those warning that make the CI fail
EDIT2: maybe fixed (not sure if fix is correct though)

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1466,14 +1466,14 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
                         (rust--format-get-loc buffer start)
                         (rust--format-get-loc buffer point))
                   window-loc)))))
-    (defvar w-start)
     (unwind-protect
         ;; save and restore window start position
         ;; after reformatting
         ;; to avoid the disturbing scrolling
-        (setq w-start (window-start))
-      (rust--format-call (current-buffer))
-      (set-window-start (selected-window) w-start)
+        (let ((w-start (window-start)))
+          (rust--format-call (current-buffer))
+          (set-window-start (selected-window) w-start)
+          )
       (dolist (loc buffer-loc)
         (let* ((buffer (pop loc))
                (pos (rust--format-get-pos buffer (pop loc))))
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1466,12 +1466,12 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
                         (rust--format-get-loc buffer start)
                         (rust--format-get-loc buffer point))
                   window-loc)))))
+    (defvar w-start)
     (unwind-protect
         ;; save and restore window start position
         ;; after reformatting
         ;; to avoid the disturbing scrolling
-        (defvar w-start)
-      (setq w-start (window-start))
+        (setq w-start (window-start))
       (rust--format-call (current-buffer))
       (set-window-start (selected-window) w-start)
       (dolist (loc buffer-loc)
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1467,7 +1467,13 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
                         (rust--format-get-loc buffer point))
                   window-loc)))))
     (unwind-protect
-        (rust--format-call (current-buffer))
+        ;; save and restore window start position
+        ;; after reformatting
+        ;; to avoid the disturbing scrolling
+        (defvar w-start)
+      (setq w-start (window-start))
+      (rust--format-call (current-buffer))
+      (set-window-start (selected-window) w-start)
       (dolist (loc buffer-loc)
         (let* ((buffer (pop loc))
                (pos (rust--format-get-pos buffer (pop loc))))


jjwest:
Works nicely for me. I'd suggest using a let binding instead of defvar, which creates a persistent global variable:

    (let ((w-start (window-start)))
     .....
    )

apiraino:
@jjwest thanks, appreciated! I've pushed a patch to reflect your suggestion.

Dushistov:
Issue #256 looks like also related

drrlvn:
Isn't this is the reason that `save-excursion` exists? I think it should be possible to just use:

```elisp
(save-excursion
    (rust--format-call (current-buffer)))
```

sebastiencs:
@drrlvn `save-excursion` uses markers that might be deleted by `rust--format-call`, see https://emacs.stackexchange.com/a/7577/15635

drrlvn:
@sebastiencs I see.

I tried looking at how `clang-format` does this as I've not had this issue in C++ code and it looks like the cursor position is passed as an argument and then the json output contains the location the cursor should be moved to. See [the source](https://github.com/sonatard/clang-format/blob/57a638fa6684fa036bbc83dfd845171b8aef7001/clang-format.el#L54).

Is there such an option in `rustfmt`, and if not would it maybe be possible to add it? I imagine every editor would need something like this.

apiraino:
review @nikomatsakis? anyone? suggestions?

thanks

brotzeit:
FYI it seems I was able to fix this issue by not using stdin in the `before-save-hook`. Since the buffer gets saved anyway it's no problem to run rustfmt on the files and revert the buffers afterwards. I think this should also work for `rust-mode`, but I'm not sure.

nikomatsakis:
This looks like a pretty simple patch :)

apiraino:
@nikomatsakis I'm sorry I had lost track of this issue :-) you mean this patch can be pushed forward? I think I've been using it since.

pnkfelix:
Ill try to take a look at this this week. 

It would be interesting to try to figure out how to make a regression test for this ...

nikomatsakis:
@apiraino seems ok to me to merge in this case. I don't really know how to judge if it will work, but I take your word for it. =) Would of course be nice to have a test, as @pnkfelix says.

apiraino:
@nikomatsakis thanks for having a look at this patch. I can verify the effect of the patch on a sample Rust source file. However I'm afraid I can't provide an elisp test because my knowledge on that is pretty limited. If anyone is willing to help, I'd be really happy to provide a test case for the regression.

In any case here's how to reproduce:
1) Ensure you have installed `rust-mode` from melpa (currently `rust-mode-20181008.1628.el`)
2) Download this [gist](https://gist.github.com/apiraino/aadb133a9d321d06fc14abff012b0a7f)
3) Open the downloaded file in emacs
4) Follow the instructions in the Rust test file
5) You should experience that the buffer moves at the end of the current window

With my patch step (5) does not happen

Hope this helps

brotzeit:
@apiraino You have to install rust and rustfmt in the travis config first. Then you can start with a simple test that formats a buffer. Afterwards you can write the regression test with `save-some-buffers`.

apiraino:
@brotzeit sorry for the long hiatus, got a lot going on.

So, I did a little reseach and tried writing a test for this PR. I believe what I need to test is the `window-start` value before and after running `rust-format-buffer`.

I tried allocating a temporary buffer as in other tests (`with-temp-buffer`) but `window-start` always return nil. I'm not sure I'm on the right path.
How should the test work in order to get meaningful values for `window-start`?

Here's a gist of what I am at this moment:
https://gist.github.com/apiraino/4c8abac8d23398b2bd6dcd5d3d54e1df

nikomatsakis:
Sorry for long hiatus. I'm kind of overwhelmed with review requests and haven't really had time to try this out. 

Maybe @mookid or @brotzeit can review?

mookid:
Hi @apiraino, 
Looks good.

About the test, the problem is with the `let`: 


```list

(defun rust-test-fn (code expected-result)
  (with-temp-buffer
    (rust-mode)
    (insert code)
    (message "%S" (selected-window))
    ;; check window before formatting
    (let ((w-start (window-start)))
      (message "before: %s" w-start))
    (rust-format-buffer)
    (should (equal expected-result (buffer-string)))

    ;; check window start after formatting
    (let ((w-start (window-start)))
      (message "%s" w-start))))
```

truchi:
Hi,

Any news on this ?

Thank you !

mookid:
Integrated, thanks. Sorry for the long delay! :)

PR #272, opened 2018-06-07
Fix font-locking of "let ref"

### LINKED ISSUE #271: `let ref` not highlighted
As the title says, take this example:

```rust
fn main() {
    let highlighted = ();
    let ref not_highlighted = ();
}
```

screen shot 2018-06-07 at 16 30 54

I'm using the latest version from MELPA.

tromey:
Fixes #271


tromey:
That failure doesn't seem to have anything to do with this patch.

PR #284, opened 2018-10-08
adjust docstring formatting

riscy:
Frequent user of this package -- thanks a lot.

Here are a few adjustments I made to the docstrings that might be worth merging.  In many cases the appearance of the documentation in `M-x apropos` is improved, and in the cases where I de-indented comments, the appearance of the documentation in `M-x help` is improved.

I used https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html and the built-in Emacs syntax checking as guides.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -55,13 +55,14 @@
 	  "\\_>"))
 
 (defun rust-looking-back-str (str)
-  "Like `looking-back' but for fixed strings rather than regexps (so that it's not so slow)"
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Seems good =)

tromey:
Thanks for doing this.  BTW if you don't know you can also try `M-x checkdoc`.  IIRC sometimes it's a bit too pedantic, but occasionally worth checking out.

PR #290, opened 2018-12-17
fix font-locking for dependent traits

areux:
and type parameters with trait bounds etc.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -679,11 +679,11 @@ Returns nil if the point is not within a Rust string."
      (,(concat (rust-re-grab (concat rust-re-ident "!")) "[({[:space:][]")
       1 font-lock-preprocessor-face)
 
-     ;; Field names like `foo:`, highlight excluding the :
-     (,(concat (rust-re-grab rust-re-ident) ":[^:]") 1 font-lock-variable-name-face)
-
      ;; CamelCase Means Type Or Constructor
      (,rust-re-type-or-constructor 1 font-lock-type-face)
+
+     ;; Field names like `foo:`, highlight excluding the :
+     (,(concat (rust-re-grab rust-re-ident) ":[^:]") 1 font-lock-variable-name-face)
 
      ;; Type-inferred binding
      (,(concat "\\_<\\(?:let\\s-+ref\\|let\\|ref\\)\\s-+\\(?:mut\\s-+\\)?" (rust-re-grab rust-re-ident) "\\_>") 1 font-lock-variable-name-face)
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Can you give a quick example of the cases this improves? Seems harmless enough though

areux:
2018-12-18-211352_1018x369_scrot

nikomatsakis:
Great! I see that there are some existing tests in `rust-mode-tests.el` for font-lock stuff (e.g., `font-lock-multi-raw-strings-in-a-row`) -- can you add a test in there for your new behavior?

damiencollard:
If the goal of this patch is to highlight the thing being defined and the thing being used/referenced (`Subtrait` vs `SuperTrait`, `Item` vs `BufMut`), then I'm strongly against it, as it reduces the possibilities for theming.

And, e.g. as someone who prefers having just a handful of colors rather than colors all over the place, keeping these separate allows me to make a theme to my liking: I'd typically *not* highlight the used trait.

areux:
> Great! I see that there are some existing tests in `rust-mode-tests.el` for font-lock stuff (e.g., `font-lock-multi-raw-strings-in-a-row`) -- can you add a test in there for your new behavior?

Sorry - I don't know how to write a test.

> If the goal of this patch is to highlight the thing being defined and the thing being used/referenced (`Subtrait` vs `SuperTrait`, `Item` vs `BufMut`) the same, then I'm strongly against it, as it reduces the possibilities for theming.
> 
> And, e.g. as someone who prefers having just a handful of colors rather than colors all over the place, keeping these separate allows me to make a theme to my liking: I'd typically _not_ highlight the used trait.

The goal was to enforce a simple convention by not highlighting **types** and **variables** with the same color:

> In general, Rust tends to use CamelCase for "type-level" constructs (types and traits) and snake_case for "value-level" constructs

2019-02-10-140901_893x262_scrot

but, having a hideous rainbow of confusion over a handful of meaningful colors is what most people prefer.

damiencollard:
> The goal was to enforce a simple convention by not highlighting **types** and **variables** with the same color:

I do agree with this, of course.

My impression from your initial example was that it would force both `Foo` and `Sized` in `trait Foo: Sized` to be highlighted the same. If not (and if at all possible, I don't know), it's good to me.

nikomatsakis:
OK, I've read the PR. I .. agree that the "variable" face doesn't seem right here, although I think that it would be nice to distinguish *definitions* from *references* in general (hence to color `Foo` distinct from `Bar` in `type Foo: Bar`).

nikomatsakis:
As far as writing a test, take a look at this:

https://github.com/rust-lang/rust-mode/blob/9e8b23fc7b3c42855732cc228b43ac4640521f5e/rust-mode-tests.el#L390-L393

I think you would want to do something similar, sort of like:

```
(ert-deftest font-lock-raw-string-after-normal-string-ending-in-r ()
  (rust-test-font-lock
   "type Foo: Bar;"
   '("Foo" font-lock-type-face "Bar" font-lock-type-face)))
```

nikomatsakis:
I'm going to close this for inactivity -- feel free to re-open with a test!

cc @mookid -- feel free to re-open if you prefer :) I'm just trying to close out my list of assigned PRs.

PR #291, opened 2018-12-17
Add auto pairing support for `|' in elec-pair.

### LINKED ISSUE #285: No autopair on closure argument lists "| |"
I'm using `autopair` package for electric pairs and was looking into getting the vertical pipes for closures to behave similarly. I gather this has something to do with the the mode's syntax tables, but I'm pretty ignorant of how this works in emacs.

mookid:
When using the package elec-pair, the matching `|' gets inserted like
for other parenthesis-like characters.

fix #285

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -21,6 +21,7 @@
 (require 'json)
 
 (defvar electric-pair-inhibit-predicate)
+(defvar electric-pair-pairs)
 (defvar electric-indent-chars)
 
 (defvar rust-buffer-project)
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Hmm, but this is not always desirable, right? e.g., you may want to write `a | b`?

nikomatsakis:
Also, there are some existing tests in `rust-mode-tests.el` for electric pair compatibility, can you add some tests in there for your new behavior?

mookid:
This is busted by design, given the need for the binary |.

PR #292, opened 2018-12-18
imenu: handle lifetimes in impl blocks.

### LINKED ISSUE #277: imenu ignores impls with lifetimes
Specifically, it ignores `impl`s with rust 2015 style lifetimes, but finds `impl`s with rust 2018 lifetime elision.

```
impl<'a> WithLifetime2015<'a> {
    fn foo() {}
}

impl WithLifetime2018<'a> {
    fn bar() {}
}

```

produces the following `imenu--index-alist`:

```
(("Impl"
  ("WithLifetime2018" . #<marker at 52 in tmp.rs>))
 ("Fn"
  ("foo" . #<marker at 33 in tmp.rs>)
  ("bar" . #<marker at 80 in tmp.rs>)))
```

Using rust-mode 20180626.1512.

mookid:
fix #277

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -3061,6 +3061,20 @@ extern \"rust-intrinsic\" fn five() {
       "four"
       "five"))))
 
+(ert-deftest rust-test-imenu-impl-with-lifetime ()
+  (test-imenu
+   "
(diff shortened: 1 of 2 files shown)


pnkfelix:
Thanks for the patch!

PR #293, opened 2018-12-21
Add rust-run and rust-test interactive function

TristanCacqueray:
This change adds a couple of useful commands similar to the
rust-build function.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1502,6 +1502,16 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
   (interactive)
   (compile "cargo build"))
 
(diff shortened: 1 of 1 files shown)


pnkfelix:
Hey @nikomatsakis and any other users of this package: Can you think of any reason not to merge this? E.g. is there any issue with someone else also defining their own `rust-run` etc? I do not personally think there is much risk associated with that, since these seem like you would only use them as user interactive commands, not as functions you'd call from other elisp code, but maybe I'm overlooking something?

pnkfelix:
I'm going to ahead and merge this; if there's problems we can back it out later.

PR #296, opened 2019-01-15
Add rust-indent-return-type-to-arguments defcustom variable

dekellum:
This adds a user customizable setting which defaults (t) to the current behavior for `-> ReturnType` on its own line, which is indenting to the same column as the arguments of the associated function. 

When disabled (nil) by the user, however, this will now just indent such a line one level from its baseline (no special treatment).

For example, the existing and default indentation:
```rust
pub fn fetch<B: hyper::body::Payload>(rr: RequestRecord<B>, tune: &Tunables)
                                      -> Result<Dialog, Flare>
{
    //...
}
```

With this change, and disabled (set nil) by the user:
```rust
pub fn fetch<B: hyper::body::Payload>(rr: RequestRecord<B>, tune: &Tunables)
    -> Result<Dialog, Flare>
{
    //...
}
```
The former appears to accommodate a visual-indent style for functions, and I suspect it predates the rustfmt and fmt-RFCs moving firmly away from visual-indent style? The fmt-RFCs also don't advertise the desirability of a line-break before `->`, but some existing code, including my own, finds a line-break here convenient. 

Includes a basic test.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -450,6 +450,17 @@ fn foo4(a:i32,
 }
 "))
 
+(ert-deftest indent-return-type-non-visual ()
+  (let ((rust-indent-return-type-to-arguments nil))
+(test-indent
+   "
+fn imagine_long_enough_to_wrap_at_arrow(a:i32, b:char)
+    -> i32
+{
+    let body;
+}
+")))
+
 (ert-deftest indent-body-after-where ()
   (let ((rust-indent-where-clause t))
     (test-indent
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -131,6 +131,13 @@ When nil, `where' will be aligned with `fn' or `trait'."
   :group 'rust-mode
   :safe #'booleanp)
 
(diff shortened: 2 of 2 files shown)


dekellum:
Note, the new non-default indentation (2nd example) as provided in this PR, appears to be what vim does by default. For emacs, it seems it should be made a non-default option (as per changes) before considering a change to the default.

tromey:
I didn't look at this patch at all; but my view overall is that the Emacs mode should agree with `rustfmt` by default.  It sounds like maybe this adds such a change but doesn't make it the default?

dekellum:
Correct, this PR doesn't change the default behavior, but provides a new non-default option which works better with some existing projects: my own, and some more prominent!  At least until rustfmt is fully integrated/automated for these projects, local convention remain important.

The case is a gray area, because writing `-> ReturnType` on its own line isn't something rustfmt will currently do, but emacs rust-mode (or vim, other editors) must offer some indent for lines as given.

PR #297, opened 2019-01-21
Do not allow subword-match when detecting built-in formatting macros

dato:
Before this change, a user-created macro like ‘cprintln!’ would see the
‘println’ subword highlighted as a built-in macro, and the leading ‘c’
left without color.

After this commit, only full words will be recognized as built-ins, and
‘cprintln!’ is highlighted as a regular macro in full, not as a built-in.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -664,7 +664,7 @@ Returns nil if the point is not within a Rust string."
       1 font-lock-preprocessor-face keep)
 
      ;; Builtin formatting macros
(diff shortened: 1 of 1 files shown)


pnkfelix:
Is it possible to add regression tests for this to `rust-mode-tests.el` as well?

dato:
I would have liked to, but I haven’t had time in the past three months, so I don't think it's going to happen from me, sorry.

mookid:
Integrated to master, thanks!

PR #300, opened 2019-02-28
Avoid signaling "Beginning of buffer" in rust-looking-back-macro

mmilenko:
Also save some work if we don't have at least two characters behind us
in the buffer, we know we can't be looking at a macro then. The signal
occurs when one tries 'racer-describe' when positioned over "String",
which eventually ends up calling

(with-temp-buffer
  (insert "<Target=str>")
  (delay-mode-hooks (rust-mode))
  (font-lock-ensure))

This simpler block will also trigger the error

(with-temp-buffer
  (insert "<>")
  (rust-mode)
  (font-lock-ensure))

(font-lock-ensure) call ends up calling (rust-looking-back-macro).

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1234,7 +1234,7 @@ fn test4();")
   (with-temp-buffer
     (rust-mode)
     (insert str)
-    (font-lock-ensure)
+    (font-lock-fontify-buffer)
     (buffer-string)))
 
 (defun rust-test-group-str-by-face (str)
@@ -1503,7 +1503,7 @@ this_is_not_a_string();)"
 1......................500......................50
 \"#;
 ")
-    (font-lock-ensure)
+    (font-lock-fontify-buffer)
     (goto-char 530)
     (insert "#")
     ;; We have now closed the raw string.  Check that the whole string is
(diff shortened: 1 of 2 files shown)


mmilenko:
Not that it's terribly important, but the actual racer.el code that ends up stepping on this is in racer--syntax-highlight

    (with-temp-buffer
      (insert str)
      (delay-mode-hooks (rust-mode))
      (if (fboundp 'font-lock-ensure)
          (font-lock-ensure)
        (with-no-warnings
          (font-lock-fontify-buffer)))
      (setq result (buffer-string)))

when str is "<...>".

The code in the original comment is the minimal example that reproduces it.

pnkfelix:
Hey great!

Can we add a unit test for this?

mmilenko:
Hi, yeah sure, as soon as I figure out how to do it. 

Looking at rust-mode-tests.el I see a lot of calls to font-lock-fontify-buffer, and I noticed that font-lock documentation says that that's only for interactive use and font-lock-ensure should be used instead from lisp code. If there's no objection I'll go ahead and make the switch throughout.

mmilenko:
Well, I'll try to do it and see what breaks and why.

mmilenko:
Done adding the unit test.

PR #303, opened 2019-04-15
Fix raw string propertization.

### LINKED ISSUE #270: (right-word) gets confused on strings that end with "r"
Is anyone else seeing this? Put the following string into a buffer, put point on the left column (on the "f") and then use right-word (aka C-right) to jump to the end of each string:

```
foo("error", "erroq");
```

On my system, in rust-mode, it stops on the open-parenthesis, on the last "r" in "error", and then on the closing quote after the "q". In any other mode the second stop is on the closing quote after "error", not on the final "r".

I'm guessing that the syntax table is a bit confused, and that it sees `r"` as the beginning of a raw string (rather than the end of a normal string).

I'm using FSF emacs-25.3.1 on OS-X (hand compiled, not Homebrew), and the ELPA packaged version of rust-mode, whose `rust-mode.el` matches the contents of this repo as of the latest commit to master (07e16c6364bf0362f76c6aa46604453dec3ac640, from 22-mar-2018).

thanks!

kurnevsky:
First, `str-start` points to quote, so we should look back for `r#`.
Second, we should put "|" property only if it's a beginning of a string.

Fixes #270

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1115,9 +1115,12 @@ raw string, or to END, whichever comes first."
     (when str-start
       (when (save-excursion
 	      (goto-char str-start)
-	      (looking-at "r\\(#*\\)\\(\"\\)"))
+	      (looking-back "r\\(#*\\)" nil))
 	;; In a raw string, so try to find the end.
-	(let ((hashes (match-string 1)))
+	(let* ((hashes (match-string 1))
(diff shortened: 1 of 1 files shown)


kurnevsky:
Ah, nevermind. I didn't understand it correctly.

PR #304, opened 2019-04-15
Don't insert string delimiter inside strings.

### LINKED ISSUE #227: rust-beginning-of-defun for function with subfunction
Similar to #222  :

```rust
    fn it_works() {
            fn helper() {}
           let a = 1 + 1;<CURSOR_HERE>
    }
```

If place cursor at marked place and execute `M-x rust-beginning-of-defun` 
then cursor jump to begin of `fn helper()` instead of `fn it_works`.

May be cause kwrooijen/cargo.el#41 (see last comment)

### LINKED ISSUE #270: (right-word) gets confused on strings that end with "r"

kurnevsky:
Check if we are inside string before inserting string delimiter.

Fixes #270

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1452,6 +1452,19 @@ fn g() {
      "g" font-lock-function-name-face
      "\"xs\"" font-lock-string-face)))
 
+(ert-deftest font-lock-string-ending-with-r-word-boundary ()
+  (with-temp-buffer
+    (rust-mode)
+    (insert "const foo = \"foo bar\"")
+    (font-lock-fontify-buffer)
+    ;; right-word should move the point to the end of the words.
+    (goto-char 14)
+    (right-word)
+    (should (equal 17 (point)))
+    (right-word)
+    (should (equal 21 (point)))
+    ))
+
 (ert-deftest font-lock-raw-string-trick-ending-followed-by-string-with-quote ()
   (rust-test-font-lock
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Do you think you can add a test of some kind?

nikomatsakis:
I'm not 100% sure how one would write such a test mind you

kurnevsky:
Added a tests for word boundaries - it fails before the fix and succeeds after.

nikomatsakis:
Thanks @kurnevsky! Sorry for the delay in reviewing.

kurnevsky:
No problem. I know you are busy :)

PR #309, opened 2019-05-28
Add rust-format-diff-buffer

phillord:
This command displays the output of rustfmt-diff.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1471,7 +1471,7 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
      (current-buffer))
    :sentinel 'rust-format-diff-buffer-sentinel))
 
-(defun rust-format-diff-buffer-sentinel (process e)
+(defun rust-format-diff-buffer-sentinel (process _e)
   (when (eq 'exit (process-status process))
     (if (> (process-exit-status process) 0)
         (display-buffer "*rustfmt-diff*")
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1456,6 +1456,27 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
             (forward-char columns)))
         (min (point) max-pos)))))
 
+(defun rust-format-diff-buffer ()
+  "Show diff to current buffer from rustfmt."
+  (interactive)
+  (unless (executable-find rust-rustfmt-bin)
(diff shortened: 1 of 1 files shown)


phillord:
Hmmm. Emacs-24. I can make this work with Emacs-24, but given that we are near the end of the Emacs-26 release cycle do you really want that support?

nikomatsakis:
A good question. All of my machines are emacs 25 or above. @pnkfelix -- thoughts?

phillord:
If you want emacs-24, I can add another PR which allows the tests to be run locally via docker. I still support Emacs-24 in one of my packages (cause it's a test framework), but it's very hard to build all of the Emacs-24 release cycle now.

nikomatsakis:
@phillord I personally don't think it's necessary, but I guess I have no real sense of what is the norm in terms of backwards compatibility for emacs packages.

phillord:
Explicit support for Emacs-24 is pretty rare now, although many packages will support it just because they do not depend on anything from 25 onwards. There are some people who are still on Emacs-24, perhaps because of organisational policy. 

I am not sure it's a question of the norm, though. it a question of Rust developers. It wasn't that long ago that most Rust developers were on nightly (perhaps they still are). And Emacs-25 was released at about the same time as Rust 1.12, while Emacs-24.5 was released about the same time as Rust 1.0.

Anyway, how do we move this forward? It's not my decision as to which version you want to support, but it would be nice to close this PR (supporting 24 or not).

nikomatsakis:
> Anyway, how do we move this forward? It's not my decision as to which version you want to support, but it would be nice to close this PR (supporting 24 or not).

Indeed. I vote we drop emacs 24 support, but I'm going to defer to @mookid, since they recently volunteered to help keep up maintenance. =)

mookid:
Hi @phillord,

is that possible to use create-process instead of make-process? If that makes compilation happy, we can defer the discussion about maintaining compatibility for emacs-24.

mookid:
I rebased on master, and added a couple commits to
- fix the build on emacs24
- use diff-mode in the resulting buffer, making ir also read-only :
https://github.com/rust-lang/rust-mode/tree/rust-format-diff-buffer


The motivation for the second point is to have something similar to vc-diff (the window get focus, so that 'q' exist).

@phillord let me know what you think.


Please let me

mookid:
Many thanks for the contribution.

PR #310, opened 2019-06-01
Add scoped pub support to beginning-of-defun

czipperz:
This now supports `pub(a::b::c) unsafe (fn|struct|...)` instead of only `pub (fn|struct|...)`.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -1094,6 +1094,25 @@ fn test4();")
       (beginning-of-defun 2)
       (should (eq (point) fn-1)))))
 
+(ert-deftest rust-beginning-of-defun-pub-scoped ()
+  (let (fn-1-start fn-1-end fn-2-start fn-2-end)
+   (with-temp-buffer
+     (rust-mode)
+     (setq fn-1-start (point))
+     (insert "pub(crate::mod1) fn test2() {}\n")
+     (setq fn-1-end (point))
+     (setq fn-2-start (point))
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -51,6 +51,8 @@
 ;;; Start of a Rust item
 (defvar rust-top-item-beg-re
   (concat "\\s-*\\(?:priv\\|pub\\)?\\s-*"
+          "\\(:?([[:word:]_[:digit:]:]+)\\)?\\s-*"
+          "\\(:?unsafe\\)?\\s-*"
```
jimblandy, line 55: Are these supposed to be shy groups? The syntax there is `?:`, not `:?`.

Rather than using `[[:word:]_[:digit:]:]+`, I think it would be better to say `\_+(::\_+)*`. This uses the `\_` "symbol constituent" syntax class.

jimblandy (changes requested):
Looks good, except for the problem with the shy groups. Just want to make sure I know what's going on there.

rust-mode.el
jimblandy: Uhh, I got the syntax for symbol constituents wrong. But this should probably use `rust-re-ident`.

jimblandy:
Most of this PR has been superseded by #346.

jimblandy:
Rebased and isolated the tests, which are the only thing that other PRs haven't covered.

PR #313, opened 2019-07-06
Speed up line indentation inside strings

riscy:
- `(line-number-at-pos)` uses `count-lines`.  The latter performs a linear scan of the entire buffer in elisp by stepping 40 lines at a time.  This means the lower you are in the buffer, the longer this function takes.  It is noticeable in large files.  This commit avoids this by introducing `rust--same-line-p`.
- `(save-excursion (forward-line -1) (end-of-line) (point))` is equivalent to `(line-end-position 0)`

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -529,8 +529,9 @@ buffer."
 
 (defun rust--same-line-p (pos1 pos2)
   "Return non-nil if POS1 and POS2 are on the same line."
-  (= (save-excursion (goto-char pos1) (line-end-position))
-     (save-excursion (goto-char pos2) (line-end-position))))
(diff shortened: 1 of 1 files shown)


mookid:
> Yeah -- there's this
> 
> ```emacs-lisp
> (defun rust--same-line-p (pos1 pos2)
>   "Return non-nil if POS1 and POS2 are on the same line."
>   (save-excursion (= (progn (goto-char pos1) (line-end-position))
>                      (progn (goto-char pos2) (line-end-position)))))
> ```
> 
> as one alternative.

ok to merge with that change applied.

riscy:
All set, thanks!

PR #314, opened 2019-07-07
integrate rustfmt error messages

nikodemus:
I found myself irked by rustfmt detected errors being hidden. This makes life better in my opinion.

Changing the default behaviour may be controversial, though...

```
   If rustfmt reports an error, by default display it in the message
   area and jump to the location. (custom rust-format-goto-problem)

   Optionally display the *rustfmt* buffer when rustfmt reports an error.
   (custom rust-format-show-buffer)
```

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -160,6 +160,18 @@ to the function arguments.  When nil, `->' will be indented one level."
   :safe #'booleanp
   :group 'rust-mode)
 
+(defcustom rust-format-show-buffer nil
+  "Show *rustfmt* buffer if formatting detected problems."
+  :type 'boolean
+  :safe #'booleanp
+  :group 'rust-mode)
+
+(defcustom rust-format-goto-problem t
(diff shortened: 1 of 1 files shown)


brotzeit:
I solved this by using a compilation mode for rustfmt.

rust-mode.el
```diff
@@ -1528,6 +1606,7 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 (defvar rust-mode-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-c C-f") 'rust-format-buffer)
+    (define-key map (kbd "C-c C-n") 'rust-goto-format-problem)
```
mookid: rather than adding a custom binding, is that possible to interface rust-goto-format-problem with next-error?

mookid:
Integrated.  Thanks for the great patch!

PR #319, opened 2019-08-31
Feature/update emacs versions

phillord:
Following on from the discussion here:

https://github.com/rust-lang/rust-mode/pull/309

This adds a updated test environment. A local docker based test is added, and all versions 25 and 26 versions are supported.


nikomatsakis:
Looks great!

PR #326, opened 2019-10-02
Highlight variable name in a for-loop

Hi-Angel:


rust-mode.el
```diff
@@ -711,6 +711,9 @@ Returns nil if the point is not within a Rust string."
      ;; Lifetimes like `'foo`
      (,(concat "'" (rust-re-grab rust-re-ident) "[^']") 1 font-lock-variable-name-face)
 
+     ;; variable name in for-loop
+     (,(concat "\\bfor[[:space:]]+" (rust-re-grab rust-re-ident) "[[:space:]]+in\\b") 1 font-lock-variable-name-face)
```
mookid: The effect of that is that the highlighting of the identifier will not start before `in` has been inputted. Is this voluntary? If so, please add a test for that, otherwise, you can simplify the regex.

Hi-Angel: @mookid yeah, I'm afraid of false-positives. Otherwise in something like this `for Ok(x) in expr()` the `Ok` would be highlighted as a variable.

Hi-Angel: *(though this won't compile, but I think there's a way to pattern-match something in for-loop)*.

I'll add an additional testcase.

mookid: > @mookid yeah, I'm afraid of false-positives. Otherwise in something like this `for Ok(x) in expr()` the `Ok` would be highlighted as a variable.

not sure about that, I got 
image

(which is not ideal, but not what you describe either) in any case.

Hi-Angel: Hmm, okay… Yeah, thanks, I'll look then into simplifying the regex, and will add an additional testcase either way.

Hi-Angel: @mookid thanks, comments addressed: I did both: simplified the regex and added another testcase.

As a side note: the variable font lock inside pattern-match, i.e. the `var` in `Foo{var}` probably have to be implemented somewhere separately as it isn't specific to a for-loop.

Hi-Angel: Hmm… Please wait a bit, I thought about something… I think, in this regex shape I could actually just add `for` to the existing font-lock pattern, which is

```lisp
(,(concat "\\_<\\(?:let\\s-+ref\\|let\\|ref\\)\\s-+\\(?:mut\\s-+\\)?" (rust-re-grab rust-re-ident) "\\_>") 1 font-lock-variable-name-face)
```

let me try

Hi-Angel: Okay, yeah, this seems to work as expected. Done.

mookid:
Looks good. I give you a couple more days to amend/refine the change before merging.

Hi-Angel:
Okay, thanks! I'll probably gonna code some more in Rust this weekend — gonna train myself by solving some puzzles on hackerrank.

Issue #302, opened 2019-03-29
Rustc error containing dash doesn't match `compilation-mode` regexp


snizovtsev:
`compile.el` regexps fails to highlight link to the source code if error message contains dash.
This happens because `rust-mode` wrongly assumes that there is no dashes before `-->` marker:

https://github.com/rust-lang/rust-mode/blob/a871d108c745276e93ff1dec7c65e2e5a57e6d8c/rust-mode.el#L1609-L1616

So in this `compilation-mode` buffer hyperlink to the `main.rs` is broken:
```
error[E0277]: expected a `std::ops::Fn<()>` closure, found `std::result::Result<fn(gotham::state::State) -> (gotham::state::State, &'static str) {say_hello}, _>`
  --> src/main.rs:12:5
   |
12 |     gotham::start(bind_addr, Ok(say_hello));
   |     ^^^^^^^^^^^^^ expected an `Fn<()>` closure, found `std::result::Result<fn(gotham::state::State) -> (gotham::state::State, &'static str) {say_hello}, _>`
   |
   = help: the trait `std::ops::Fn<()>` is not implemented for `std::result::Result<fn(gotham::state::State) -> (gotham::state::State, &'static str) {say_hello}, _>`
   = note: wrap the `std::result::Result<fn(gotham::state::State) -> (gotham::state::State, &'static str) {say_hello}, _>` in a closure with no arguments: `|| { /* code */ }
   = note: required because of the requirements on the impl of `gotham::handler::NewHandler` for `std::result::Result<fn(gotham::state::State) -> (gotham::state::State, &'static str) {say_hello}, _>`
   = note: required by `gotham::start`
```
We can explore `compilation-debug` text property with `M-x describe-text-property` and see that built-in [gnu rule](https://github.com/emacs-mirror/emacs/blob/273d7b3ee0f4841c7f3f112aeb2e29db51a642e7/lisp/progmodes/compile.el#L253) matches.

Ideally, as Emacs code says, all registered regexps should be tested together to avoid conflicts:
```elisp
;; If you make any changes to `compilation-error-regexp-alist-alist',
;; be sure to run the ERT test in test/lisp/progmodes/compile-tests.el.
;; emacs -batch -l compile-tests.el -f ert-run-tests-batch-and-exit
(defvar compilation-error-regexp-alist-alist
```

MageSlayer:
```diff
diff --git a/rust-mode.el b/rust-mode.el
index 4ded41a..db53c67 100644
[...]

PR #331, opened 2019-10-10
Fix rustc-compilation-regexps: match error messages with dashes.

mookid:


mookid:
cc @MageSlayer @snizovtsev 

fix #302

PR #337, opened 2019-10-26
electric-pair-mode: prevent self-insert of `>' when already inserted.

### LINKED ISSUE #333: electric-pair-mode work incorrect with `<>`
I started Emacs and opened rust file as below,

```
emacs -q
M-x package-initialize
M-x electric-pair-mode
(require 'rust-mode)
```

If I type `tmp(x)` it works correctly and results in `tmp(x)`
However if I type `tmp<T>` it would result in `tmp<T>>`

It seems like this has to do with `rust-ordinary-lt-gt-p`.
I found `tmp<T>` works fine with `(defun rust-ordinary-lt-gt-p () nil)`
(although it ruins the other bracket processes).

So I guess `rust-ordinary-lt-gt-p` should be updated to return `nil` properly in the specific case.

Please correct me if I'm missing something.

Thanks.

mookid:
Fix #333.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -3220,4 +3220,18 @@ impl Two<'a> {
 
   (ert-deftest rust-test-electric-pair-lt-expression-capitalized-keyword ()
     (test-electric-pair-insert "fn foo() -> Box" 16 ?< ?>))
-  )
+
+  (ert-deftest rust-test-electric-pair-<-> ()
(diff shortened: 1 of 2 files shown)


mookid:
@tttuuu888   can you please test the fix?

tttuuu888:
@mookid This fixes the problem and work fine for me. Thank you so much.

PR #338, opened 2019-10-29
Allow pass custom flags to rustfmt.

### LINKED ISSUE #289: Use `cargo fmt` instead of `rustfmt`
For Rust 2018, rustfmt does not work correctly (e.g. removes leading `::` from `use` statements. The fact that `rust-mode` runs it instead of `cargo fmt` makes it problematic for writing 2018 code...

It would be nice to have this mode use the recommended `cargo fmt` command instead of `rustfmt`.

Thanks in advance!

mookid:
* rust-rustfmt-switches: new defvar.
* rust-format-diff-buffer, rust--format-call: use the custom flags.

Fix #289.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -165,6 +165,11 @@ to the function arguments.  When nil, `->' will be indented one level."
   :type 'string
   :group 'rust-mode)
 
+(defcustom rust-rustfmt-switches '("--edition" "2018")
+  "Arguments to pass when invoking the `rustfmt' executable."
+  :type '(repeat string)
+  :group 'rust-mode)
+
 (defcustom rust-cargo-bin "cargo"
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -1466,15 +1477,17 @@ Return the created process."
...
-                        rust-rustfmt-bin
-                        "--check"
-                        (buffer-file-name))))
+         (apply 'start-process
+                "rustfmt-diff"
+                (with-current-buffer
+                    (get-buffer-create "*rustfmt-diff*")
+                  (let ((inhibit-read-only t))
+                    (erase-buffer))
+                  (current-buffer))
```
brotzeit: Maybe make a let binding for the buffer ?

PR #339, opened 2019-10-31
Fix rust-insert-dbg for emacs-version < 25.

mookid:
The behaviour of insert-parenthesis changed with emacs 25: 

+ point and mark are not exchanged, even if out of order
+ the point ends up before the opening parenthesis in emacs 24.5, while after it on 25.1.

Hence this fix.


mookid:
cc @r-darwish

PR #340, opened 2019-11-01
mention LSP support

brotzeit:


README.md
```diff
@@ -36,7 +43,26 @@ And put this in your config to load rust-mode automatically:
...
+       (expand-file-name "~/tmp/git/straight.el/bootstrap.el" user-emacs-directory))
+      (bootstrap-version 5))
+  (unless (file-exists-p bootstrap-file)
+    (with-current-buffer
+        (url-retrieve-synchronously
+         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
+         'silent 'inhibit-cookies)
+      (goto-char (point-max))
+      (eval-print-last-sexp)))
+  (load bootstrap-file nil 'nomessage))
```
mookid: this can be found, up to date, in https://github.com/raxod502/straight.el.

I don't think this is always correct (what if "~/tmp/" does not exist?).
Maybe you can just mention that this is an option that exists and redirect to one resource rather than putting that in here.

README.md
```diff
@@ -71,6 +97,24 @@ The file `rust-mode-tests.el` contains tests that can be run via
 You can use `run_rust_emacs_tests.sh` to run them in batch mode, if
 you set the environment variable EMACS to a program that runs emacs.
 
+# LSP
+
+## eglot
+
+```elisp
+(require 'eglot)
```
mookid: same here. even if the setup is very simple, it can end up being out of sync with upstream.

README.md
brotzeit: True.

README.md
```diff
@@ -5,16 +5,23 @@
 <!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
```
mookid: Did you run that command?  on my machine it produces the following diff:

``` diff
@@ -5,6 +5,7 @@
 <!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
 **Table of Contents**
 
+- [Emacs mode for editing Rust source code](#emacs-mode-for-editing-rust-source-code)
 - [Installation](#installation)
     - [Melpa](#melpa)
     - [straight](#straight)

```



It's not related to your change, just curious if markdown-toc-refresh-toc can be trusted at all.

brotzeit: Yeah it's a little annoying, that you have to remove the first line every time you run markdown-toc-refresh-toc. But aside from this it seems to work fine.

mookid: What I was getting at is that if its reliable, it could be added to the CI script.

PR #343, opened 2019-11-23
add support for "async fn"

haroldcarr:
Async functions do not show up.   Adding this causes them to show in the imenu.


brotzeit:
Thanks!

PR #344, opened 2019-11-27
more imenu support

haroldcarr:
Before things like "pub(super) fn" were not listed.  Now they are.

All the additions work and are tested, EXCEPT I did not test pub(in SimplePath) fn

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1292,7 +1292,17 @@ whichever comes first."
 (defvar rust-imenu-generic-expression
   (append (mapcar #'(lambda (x)
                       (list (capitalize x) (rust-re-item-def-imenu x) 1))
-                  '("async fn" "enum" "struct" "union" "type" "mod" "fn" "trait" "impl"))
+                  ;; Listed in reverse order of how they are displayed (when found).
+                  '(
+                    "fn"
+                    "async\sfn"
+                    "pub\sasync\sfn"
+                    "pub\sfn"
+                    "pub\s*(\s*crate\s*)\s*fn"
+                    "pub\s*(\s*self\s*)\s*fn"
+                    "pub\s*(\s*super\s*)\s*fn"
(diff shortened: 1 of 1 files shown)


mookid:
Great! I will take a look at the code this weekend :-)

rust-mode.el
```diff
@@ -1292,7 +1292,17 @@ whichever comes first."
...
   (append (mapcar #'(lambda (x)
                       (list (capitalize x) (rust-re-item-def-imenu x) 1))
-                  '("async fn" "enum" "struct" "union" "type" "mod" "fn" "trait" "impl"))
+                  ;; Listed in reverse order of how they are displayed (when found).
+                  '(
+                    "fn"
+                    "async\sfn"
+                    "pub\sasync\sfn"
+                    "pub\sfn"
+                    "pub\s*(\s*crate\s*)\s*fn"
```
mookid: do we really want to have, pub(fn) and fn as different categories? It seems a bt counter-intuitive to me.

mookid:
Thanks for the pointer to the missing items!

Reading from the definition of `imenu-generic-expression`, it seems that what you are trying to achieve could be best done extending the pattern to match an optional pub(super|self|crate) prefix without adding categories. What do you think?

rust-mode.el
haroldcarr: These are hard questions, since many people probably have different preferences.

That said, good software engineering suggests making a public API and an private implementation.   Having imenu separate these too seems to follow.

haroldcarr: Matching pub(super|crate ... seems like a good idea.

mookid: what about pub(crate) vs pub(super) vs any other granularity level?  I would guess that the distinction between function vs enum vs trait is more meaningful that the one between the different visibility levels of fn.

mookid: cc @brotzeit.

mookid:
thanks for the input! please test.

Issue #295, opened 2019-01-08
Support new visibility keywords in imenu-generic-expression


shanavas786:
`imenu` doesn't detect functions with `crate` or `pub(in ..)` visibility

mookid:
fixed with #346. please test!

PR #346, opened 2019-12-01
imenu: fn items: match async/const and all pub variants.

mookid:
see https://doc.rust-lang.org/reference/items/functions.html and https://doc.rust-lang.org/reference/visibility-and-privacy.html for the spec of pattern to be matched.


mookid:
cc @haroldcarr 

This is my alternative proposal for #344. 

fix #295 
fix #344

Issue #208, opened 2017-04-28
Significant slowdown with rust-match-angle-brackets non nil


stianeklund:
I'm experiencing serious slowdowns when entering insert mode & navigating in general all of a sudden.
Not sure what the cause is but figured I'd create an issue as it doesn't happen with `rust-mode` turned off.

Here's the output from the profiler-report:

```

Function                            Cpu samples   %

timer-event-handler                       13844  92%
command-execute                             782   5%
 ...                                        187   1%
redisplay_internal (C function)             169   1%
eldoc-pre-command-refresh-echo-area           5   0%
sp--save-pre-command-state                    3   0%

```

If I expand `timer-event-handler`, most of the CPU is being hogged on `rust-in-macro` within the `up-list` function. Also `backward-up-list` hogs a good chunk of CPU.
```
 - apply                                                         2171  77%
  - hl-paren-highlight                                           2090  74%
   - scan-sexps                                                  2077  74%
    - internal--syntax-propertize                                2077  74%
     - syntax-propertize                                         2077  74%
      - rust-syntax-propertize                                   2073  74%
       - rust-ordinary-lt-gt-p                                   2073  74%
        - rust-in-macro                                          1955  69%
         + rust-in-macro                                         1615  57%
         + backward-up-list                                       241   8%
         + rust-rewind-irrelevant                                  95   3%
         + rust-looking-back-ident                                  4   0%
        + rust-in-str-or-cmnt                                     118   4%
```
Anyone experienced anything similar?

Same issue on current nightly & stable (1.17) I've tried with a clean .spacemacs file and without any additional packages. It's only when `rust-mode` is enabled I run into this slowdown.

The file I'm working on is less than 1k lines.

Also as an aside, I've tried Emacs24, Ubuntu 15.04, 16.04 and now upgraded to 16.10.
This is on a Virtual machine.

```
Description:	Ubuntu 16.10
Release:	16.10
Codename:	yakkety
```

tromey:
Sounds like a regression due to the syntax propertization patch; but the spacemacs part makes me wary, I don't really know spacemacs and in many ways it is its own thing.  I wonder if there's a way to reproduce without it.

stianeklund:
@tromey Now for some reason it works great without any slowdowns. I can't reproduce this issue now anymore for some weird reason. I wish I had something to bring to the table here but I do not.

I'll try upgrading my computer at home and see if I run into the same issues.

fafhrd91:
[...]

PR #349, opened 2020-01-20
change rust-match-angle-brackets default to nil

brotzeit:
#208


mookid:
@brotzeit let's try a bit more to make the current default work, thanks to #369 and #372 in particular. Once the improvements in place and if we keep hitting that kind of issue, let's reconsider.

PR #355, opened 2020-02-19
integrate rustfmt error messages

mookid:
Re #314 
I finally had some time to look at that one, I just fixed a typo on the way. Thanks for the patch, I love it!

cc @nikodemus

@brotzeit What do you think?

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1440,9 +1440,11 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
     (point)))
 
 ;; Formatting using rustfmt
+(defconst rust-rustfmt-buffername "*rustfmt*")
+
 (defun rust--format-call (buf)
   "Format BUF using rustfmt."
-  (with-current-buffer (get-buffer-create "*rustfmt*")
+  (with-current-buffer (get-buffer-create rust-rustfmt-buffername)
     (erase-buffer)
     (insert-buffer-substring buf)
     (let* ((tmpf (make-temp-file "rustfmt"))
@@ -1479,7 +1481,7 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 ;; Since we run rustfmt through stdin we get <stdin> markers in the
 ;; output. This replaces them with the buffer name instead.
 (defun rust--format-fix-rustfmt-buffer (buffer-name)
-  (with-current-buffer (get-buffer "*rustfmt*")
+  (with-current-buffer (get-buffer rust-rustfmt-buffername)
     (goto-char (point-min))
     (while (re-search-forward "--> <stdin>:")
       (replace-match (format "--> %s:" buffer-name)))))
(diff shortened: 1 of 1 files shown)


brotzeit:
I will take a look at it the next days.

@tarsius is working on a solution to making rustic modular https://github.com/brotzeit/rustic/pull/95.
Maybe we can start another try to bring rust-mode and rustic together and reduce the maintenance burden.

rust-mode.el
```diff
@@ -198,6 +198,18 @@ to the function arguments.  When nil, `->' will be indented one level."
   :safe #'booleanp
   :group 'rust-mode)
 
+(defcustom rust-format-show-buffer nil
```
brotzeit: Why not set the default to t ? I think most users want the buffer the be displayed when there's an error ?

rust-mode.el
```diff
@@ -1455,13 +1467,76 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
             (error "Rustfmt failed, see *rustfmt* buffer for details"))))
       (delete-file tmpf))))
 
+;; Since we run rustfmt through stdin we get <stdin> markers in the
+;; output. This replaces them with the buffer name instead.
+(defun rust--format-fix-rustfmt-buffer (buffer-name)
+  (with-current-buffer (get-buffer "*rustfmt*")
+    (goto-char (point-min))
+    (while (re-search-forward "--> <stdin>:")
+      (replace-match (format "--> %s:" buffer-name)))))
```
brotzeit: I think you can drop the `-->`.

rust-mode.el
```diff
@@ -1455,13 +1467,76 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
...
+(defun rust-goto-format-problem ()
+  "Jumps to problem reported by rustfmt, if any.
+
+In case of multiple problems cycles through them. Displays the
+rustfmt complain in the echo area."
+  (interactive)
+  ;; This uses position in *rustfmt* buffer to know which is the next
+  ;; error to jump to, and source: line in the buffer to figure which
+  ;; buffer it is from.
+  (let ((rustfmt (get-buffer "*rustfmt*")))
```
brotzeit: Maybe introduce a `defvar` for the buffer name ?

rust-mode.el
mookid: I think that the idea is to show the error in the echo area by default.

brotzeit: I see.

brotzeit:
FYI, using `call-process-region` in `rust--format-call` doesn't allow us to fix this kind of issue https://github.com/brotzeit/rustic/issues/19. But I guess that's not relevant for most users.

mookid:
> 
> 
> FYI, using `call-process-region` in `rust--format-call` doesn't allow us to fix this kind of issue [brotzeit/rustic#19](https://github.com/brotzeit/rustic/issues/19). But I guess that's not relevant for most users.

Thanks for the pointer. A reference to the fixing commit would also be  useful ;)

brotzeit:
I removed the note about that because I thought you might don't like the solution =)

We would have to add some more logic around a call to `make-process`. [This](https://github.com/brotzeit/rustic/blob/master/rustic-compile.el#L201) would be a solution to the problem.

And when we add this it would be silly not to implement translation of ANSI control sequences through xterm-color(or whatever). So it would end up being more additional code and we decided to keep rust-mode a lightweight package.

mookid:
> FYI, using `call-process-region` in `rust--format-call` doesn't allow us to fix this kind of issue [brotzeit/rustic#19](https://github.com/brotzeit/rustic/issues/19). But I guess that's not relevant for most users.

I can't reproduce that one. Let's keep an eye on it.

PR #359, opened 2020-03-09
Don't fail on re-search-forward.

### LINKED ISSUE #357: crippy bug: stdin buffer if "Rust Format On Save" is on
It is random, I have no idea how to reproduce it.

Time to time when I save current buffer with rust-mode, let's call it `a.rs`,
emacs does not save it, and instead it open buffer stdin, in other words I press
C-x s and current buffer change to "stdin" buffer. This buffer is empty,
and `a.rs` still in modified state. I switch back to `a.rs`, try to save it and again empty stdin instead of a.rs.

So actually I can not save my rust buffer,
I have to set "Rust Format On Save" to off to make possible save,
or change mode from rust-mode to text-mode.

This bug appear only recently, so I suppose it is related to last "rustfmt" changes in master.

GNU Emacs 26.3 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.10) of 2019-08-29
rust-mode          20200229.1547

mookid:


mookid:
@jamii @Dushistov  can you please try this fix?

jamii:
I get the same behaviour as before.

mookid:
Thanks for the help. I'll ty again in the next few days; I tend to believe that this is a change of output format from rustfmt.

Dushistov:
I can not reproduce #357 since I sync my rust-mode with master to get debug commit. Before it was reproduced once a day, now I can not reproduce it at all. But may be I need to write new code , and not edit old one as mentioned [here](https://github.com/rust-lang/rust-mode/issues/357#issuecomment-596674683). But I should say that, when I able to reproduce this bug, sometimes it shows "rustfmt" error just fine, and sometime it switch to stdin, and I couldn't find out regularity.

PR #362, opened 2020-03-23
Add rust-occur-definitions

lerouxrgd:
Taking inspiration from [elpy](https://github.com/jorgenschaefer/elpy/blob/b7396efa97c507ceb26ef479b05ec04ff382ec14/elpy.el#L2522), this function displays an occur buffer of all definitions in the current buffer.

This is similar to what `imenu` provides, but it also reflects the overall file structure (by preserving order and indentation).

It could be bound to `C-c C-o` (as done in `elpy`).

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1378,6 +1378,29 @@ Create a hierarchical index of the item definitions in a Rust file.
 Imenu will show all the enums, structs, etc. in their own subheading.
 Use idomenu (imenu with `ido-mode') for best mileage.")
 
+;;; Occur support
+
+(defun rust-occur-definitions ()
+  "Display an occur buffer of all definitions in the current buffer."
+  (interactive)
+  (let ((list-matching-lines-face nil))
+    (occur (concat "^\s*\\("
+                   "\\(pub.*?\s\\|\\)mod\\|"
+                   "\\(pub.*?\s\\|\\)type\\|"
+                   "\\(pub.*?\s\\|\\)struct\\|"
+                   "\\(pub.*?\s\\|\\)enum\\|"
+                   "\\(pub.*?\s\\|unsafe\s\\|const\s\\|async\s\\|\\)fn\\|"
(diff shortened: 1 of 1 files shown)


rust-mode.el
```diff
@@ -1378,6 +1378,29 @@ Create a hierarchical index of the item definitions in a Rust file.
...
+                   "\\(pub.*?\s\\|\\)struct\\|"
+                   "\\(pub.*?\s\\|\\)enum\\|"
+                   "\\(pub.*?\s\\|unsafe\s\\|const\s\\|async\s\\|\\)fn\\|"
+                   "\\(pub.*?\s\\|unsafe\s\\|\\)trait\\|"
+                   "\\(pub.*?\s\\|\\)const\\|"
+                   "\\(pub.*?\s\\|\\)static\sref\\|"
+                   "lazy_static\!\\|"
+                   "impl"
+                   "\\)\s")))
+  (let ((window (get-buffer-window "*Occur*")))
```
brotzeit, line 1399: Maybe use `-if-let` here ?

tromey: Wouldn't that mean adding dash as a dependency?  It's preferable IMO to not introduce dependencies.

mookid:
I am bothered by the duplication of functionality compared to the exisiting imenu support. Did you investigate if it was possible to instead improve the imenu code to obtain a similar effect?

A good thing about this is that existing packages provide a different user interface to imenu instead of the one you suggest.

rust-mode.el
brotzeit: Oops, thanks. It's almost in every project I'm working on so I forgot it's not a built-in =)

lerouxrgd:
I understand your concern. I think imenu is more aimed at dynamic interaction (filtering, narrowing down, through different interfaces as you mentioned). Here `occur` is the final interface (navigable output, file structure overview).

So feel free to close the PR if you think this is a duplication of functionalities.

PR #369, opened 2020-04-23
Re-implement rust-in-macro faster

### LINKED ISSUE #208: Significant slowdown with rust-match-angle-brackets non nil

phillord:
This function is called repeatedly by `syntax-ppss` and so needs to be
fast.  We adopt a simpler mechanism for determining whether we are in
a macro.

Fixes #208

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -106,7 +106,12 @@ symbols."
     (looking-back rust-re-ident beg-of-symbol)))
 
 (defun rust-looking-back-macro ()
-  "Non-nil if looking back at an ident followed by a !"
+  "Non-nil if looking back at an ident followed by a !
+
+This is stricter than rust syntax which allows a space between
+the ident and the ! symbol. If this space is allowed, then we
+would also need a keyword check to avoid `if !(condition)` being
+seen as a macro."
   (if (> (- (point) (point-min)) 1)
       (save-excursion
         (backward-char)
@@ -260,18 +265,90 @@ to the function arguments.  When nil, `->' will be indented one level."
         ;; Rewind until the point no longer moves
         (setq continue (/= starting (point)))))))
 
-(defun rust-in-macro ()
+(defvar-local rust-macro-scopes nil
+  "Cache for the scopes calculated by `rust-macro-scope'.
+
+This variable can be `let' bound directly or indirectly around
+`rust-macro-scope' as an optimization but should not be otherwise
+set.")
+
+(defun rust-macro-scope (start end)
+  "Return the scope of macros in the buffer.
+
+The return value is a list of (START END) positions in the
+buffer.
+
+If set START and END are optimizations which limit the return
+value to scopes which are approximately with this range."
   (save-excursion
-    (when (> (rust-paren-level) 0)
-      (backward-up-list)
-      (rust-rewind-irrelevant)
-      (or (rust-looking-back-macro)
-          (and (rust-looking-back-ident)
-               (save-excursion
-                 (backward-sexp)
-                 (rust-rewind-irrelevant)
-                 (rust-looking-back-str "macro_rules!")))
-          (rust-in-macro)))))
+    ;; need to special case macro_rules which has unique syntax
+    (let ((scope nil)
+          (start (or start (point-min)))
+          (end (or end (point-max))))
+      (goto-char start)
+      ;; if there is a start move back to the previous top level,
+      ;; as any macros before that must have closed by this time.
+      (let ((top (syntax-ppss-toplevel-pos (syntax-ppss))))
+        (when top
+          (goto-char top)))
+      (while
+          (and
+           ;; The movement below may have moved us passed end, in
+           ;; which case search-forward will error
+           (< (point) end)
+           (search-forward "!" end t))
+        (let ((pt (point)))
+          (cond
+           ;; in a string or comment is boring, move straight on
+           ((rust-in-str-or-cmnt))
+           ;; in a normal macro,
+           ((and (skip-chars-forward " \t\n\r")
+                 (memq (char-after)
+                       '(?\[ ?\( ?\{))
+                 ;; Check that we have a macro declaration after.
+                 (rust-looking-back-macro))
+            (let ((start (point)))
+              (ignore-errors (forward-list))
+              (setq scope (cons (list start (point)) scope))))
+           ;; macro_rules, why, why, why did you not use macro syntax??
+           ((save-excursion
+              ;; yuck -- last test moves point, even if it fails
+              (goto-char (- pt 1))
+              (skip-chars-backward " \t\n\r")
+              (rust-looking-back-str "macro_rules"))
+            (save-excursion
+              (when (re-search-forward "[[({]" nil t)
+                (backward-char)
+                (let ((start (point)))
+                  (ignore-errors (forward-list))
+                  (setq scope (cons (list start (point)) scope)))))))))
+      ;; Return 'empty rather than nil, to indicate a buffer with no
+      ;; macros at all.
+      (or scope 'empty))))
+
+(defun rust-in-macro (&optional start end)
+  "Return non-nil when point is within the scope of a macro.
+
+If START and END are set, minimize the buffer analysis to
+approximately this location as an optimization.
+
+Alternatively, if `rust-macro-scopes' is a list use the scope
+information in this variable. This last is an optimization and
+the caller is responsible for ensuring that the data in
+`rust-macro-scopes' is up to date."
+  (when (> (rust-paren-level) 0)
+    (let ((scopes
+           (or
+            rust-macro-scopes
+            (rust-macro-scope start end))))
+      ;; `rust-macro-scope' can return the symbol `empty' if the
+      ;; buffer has no macros at all.
+      (when (listp scopes)
+        (seq-some
+         (lambda (sc)
+           (and (>= (point) (car sc))
+                (< (point) (cadr sc))))
+         scopes)))))
 
 (defun rust-looking-at-where ()
   "Return T when looking at the \"where\" keyword."
@@ -1206,32 +1283,34 @@ whichever comes first."
 
 (defun rust-syntax-propertize (start end)
   "A `syntax-propertize-function' to apply properties from START to END."
-  (goto-char start)
-  (let ((str-start (rust-in-str-or-cmnt)))
-    (when str-start
-      (rust--syntax-propertize-raw-string str-start end)))
-  (funcall
-   (syntax-propertize-rules
-    ;; Character literals.
-    (rust--char-literal-rx (1 "\"") (2 "\""))
-    ;; Raw strings.
-    ("\\(r\\)#*\""
-     (0 (ignore
-         (goto-char (match-end 0))
-         (unless (save-excursion (nth 8 (syntax-ppss (match-beginning 0))))
-           (put-text-property (match-beginning 1) (match-end 1)
-                              'syntax-table (string-to-syntax "|"))
-           (rust--syntax-propertize-raw-string (match-beginning 0) end)))))
-    ("[<>]"
-     (0 (ignore
-         (when (save-match-data
-                 (save-excursion
-                   (goto-char (match-beginning 0))
-                   (rust-ordinary-lt-gt-p)))
-           (put-text-property (match-beginning 0) (match-end 0)
-                              'syntax-table (string-to-syntax "."))
-           (goto-char (match-end 0)))))))
-   (point) end))
+  ;; Cache all macro scopes as an optimization. See issue #208
+  (let ((rust-macro-scopes (rust-macro-scope start end)))
+    (goto-char start)
+    (let ((str-start (rust-in-str-or-cmnt)))
+      (when str-start
+        (rust--syntax-propertize-raw-string str-start end)))
+    (funcall
+     (syntax-propertize-rules
+      ;; Character literals.
+      (rust--char-literal-rx (1 "\"") (2 "\""))
+      ;; Raw strings.
+      ("\\(r\\)#*\""
+       (0 (ignore
+           (goto-char (match-end 0))
+           (unless (save-excursion (nth 8 (syntax-ppss (match-beginning 0))))
+             (put-text-property (match-beginning 1) (match-end 1)
+                                'syntax-table (string-to-syntax "|"))
+             (rust--syntax-propertize-raw-string (match-beginning 0) end)))))
+      ("[<>]"
+       (0 (ignore
+           (when (save-match-data
+                   (save-excursion
+                     (goto-char (match-beginning 0))
+                     (rust-ordinary-lt-gt-p)))
+             (put-text-property (match-beginning 0) (match-end 0)
+                                'syntax-table (string-to-syntax "."))
+             (goto-char (match-end 0)))))))
+     (point) end)))
 
 (defun rust-fill-prefix-for-comment-start (line-start)
   "Determine what to use for `fill-prefix' based on the text at LINE-START."
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -2139,8 +2139,8 @@ fn main() {
         (should (equal (scan-sexps (+ 1 close-pos) -1) open-pos))))
     (dolist (nonpar-pos nonparen-positions)
       (let ((nonpar-syntax-class (syntax-class (syntax-after nonpar-pos))))
-        (should (not (equal 4 nonpar-syntax-class)))
-        (should (not (equal 5 nonpar-syntax-class)))))))
+        (should-not (equal 4 nonpar-syntax-class))
+        (should-not (equal 5 nonpar-syntax-class))))))
 
 (ert-deftest rust-test-unmatched-single-quote-in-comment-paren-matching ()
   ;; This was a bug from the char quote handling that affected the paren
@@ -2922,6 +2922,112 @@ macro_c!{
      123 ;; macro_c <
      125 ;; macro_d >
      )))
+
+(ert-deftest rust-test-paren-matching-no-angle-brackets-in-macro-rules ()
+  (rust-test-matching-parens
+   "
+fn foo<A>(a:A) {
+    macro_rules! foo ( foo::<ignore the bracets> );
+    macro_rules! bar [ foo as Option<B> ];
+}
+
+macro_c!{
+    struct Boo<D> {}
+}"
+   '((8 10))
+   ;; Inside macros, it should not find any angle brackets, even if it normally
+   ;; would
+   '(47 ;; foo <
+     62 ;; foo >
+     107 ;; bar <
+     109 ;; bar >
+     141 ;; macro_c <
+     143 ;; macro_c >
+     )))
+
+(ert-deftest rust-test-in-macro-do-not-fail-on-unbalance ()
+  (should
+   ;; We don't care about the results here, so long as they do not error
+   (with-temp-buffer
+     (insert
+      "fn foo<A>(a:A) {
+    macro_c!{
+        struct Boo<D> {}
+")
+     (rust-mode)
+     (goto-char (point-max))
+     (syntax-ppss))))
+
+
+(ert-deftest rust-test-in-macro-no-caching ()
+  (should-not
+   (with-temp-buffer
+     (insert
+      "fn foo<A>(a:A) {
+    macro_c!{
+        struct Boo<D> {}
+")
+     (rust-mode)
+     (search-backward "macro")
+     ;; do not use the cache
+     (let ((rust-macro-scopes nil))
+       (rust-in-macro)))))
+
+(ert-deftest rust-test-in-macro-fake-cache ()
+  (should
+   (with-temp-buffer
+     (insert
+      "fn foo<A>(a:A) {
+    macro_c!{
+        struct Boo<D> {}
+")
+     (rust-mode)
+     (search-backward "macro")
+     ;; make the cache lie to make the whole buffer in scope
+     ;; we need to be at paren level 1 for this to work
+     (let ((rust-macro-scopes `((,(point-min) ,(point-max)))))
+       (rust-in-macro)))))
+
+(ert-deftest rust-test-in-macro-broken-cache ()
+  (should-error
+   (with-temp-buffer
+     (insert
+      "fn foo<A>(a:A) {
+    macro_c!{
+        struct Boo<D> {}
+")
+     (rust-mode)
+     (search-backward "Boo")
+     ;; do we use the cache at all
+     (let ((rust-macro-scopes '(I should break)))
+       (rust-in-macro)))))
+
+(ert-deftest rust-test-in-macro-nested ()
+  (should
+   (equal
+    (with-temp-buffer
+      (insert
+       "macro_rules! outer {
+    () => { vec![] };
+}")
+      (rust-mode)
+      (rust-macro-scope (point-min) (point-max)))
+    '((38 40) (20 45)))))
+
+(ert-deftest rust-test-in-macro-not-with-space ()
+  (should
+   (equal
+    (with-temp-buffer
+      (insert
+       "fn foo<T>() {
+    if !(mem::size_of::<T>() > 8) {
+        bar()
+    }
+}")
+      (rust-mode)
+      (rust-macro-scope (point-min) (point-max)))
+    'empty)))
+
 
 (ert-deftest rust-test-paren-matching-type-with-module-name ()
   (rust-test-matching-parens


mookid:
Thanks for working on this! before diving into the code, do you have a benchmark to compare the old vs the new version?

phillord:
Yep, best benchmark of all. I tried it with the new version and rust-mode feels fine, while without it feels terrible.

For something slightly more formal  I tried this:

````
(defun bench ()
  (goto-char (point-max))
  (message "%s"
           (benchmark-run 50
             (syntax-ppss-flush-cache 1)
             (syntax-ppss))))
````
With different benchmark forms, for elapsed, 10 and 50 times

````
Elapsed
new
0.204490421

old
0.497328317

benchmark-run-compiled 10
(1.3721566170000001 4 0.39234910600000106)
(4.620864861 7 0.6880264240000002)

benchmark-run 10
(1.2887137860000002 3 0.29979056699999873)
(4.579363357999999 7 0.6914120969999971)

benchmark-run 50
(6.617156954 16 1.616331799000001)
(23.228904235999998 36 3.5606590710000035)
````

So, only about 4 times faster by this metric. Their might be something else going on like the current implementation flushing the cache more often. I did look for this but could not find it happening. 

Unwinding the recursion would probably give a bit of performance as well.

**Comment by mookid** [comment]
> Unwinding the recursion would probably give a bit of performance as well.

yep, thats what I was thinking about.

**Comment by mookid** [comment]
> 
> 
> Yep, best benchmark of all. I tried it with the new version and rust-mode feels fine, while without it feels terrible.
> 
> For something slightly more formal I tried this:
> 
> ```
> (defun bench ()
>   (goto-char (point-max))
>   (message "%s"
>            (benchmark-run 50
>              (syntax-ppss-flush-cache 1)
>              (syntax-ppss))))
> ```
> 


And for which source code do the forms run?

**Comment by phillord** [comment]
So, full code:

````

(defun rust-in-macro-old ()
  (save-excursion
    (when (> (rust-paren-level) 0)
      (backward-up-list)
      (rust-rewind-irrelevant)
      (or (rust-looking-back-macro)
          (and (rust-looking-back-ident)
               (save-excursion
                 (backward-sexp)
                 (rust-rewind-irrelevant)
                 (rust-looking-back-str "macro_rules!")))
          (rust-in-macro)))))

(defun rust-in-macro-new ()
  (save-excursion
    (let ((end-bound (point)))
      (when (and
             ;; Find first ! backward
             (rust-backward-to-macro-bang)
             ;; Find next paren
             (re-search-forward "[([{]" end-bound t))
        (backward-char)
        ;; Find matching paren
        (forward-list)
        ;; if point is now passed where we started, we are in a macro
        (> (point) end-bound)))))


(defvar bench-use-old t)

(defun rust-in-macro ()
  (if bench-use-old
      (rust-in-macro-old)
    (rust-in-macro-new)))


(defun bench-old ()
  (interactive)
  (let ((bench-use-old t))
    (bench)))

(defun bench-new ()
  (interactive)
  (let ((bench-use-old nil))
    (bench)))

(defun bench ()
  (goto-char (point-max))
  (message "%s"
           (benchmark-run 20
             (syntax-ppss-flush-cache 1)
             (syntax-ppss))))
````

Run with  point "raw_vec.rs" from liballoc I get these results

````
elapse
0.378334104
1.654185709

run 1
(0.390778682 0 0.0)
(1.992712346 1 0.3988344170000033)

run-compiled 1
(0.38834057 0 0.0)
(1.651837399 0 0.0)

run 5
(1.960683374 1 0.35779243200000366)
(8.343783432 1 0.3593990110000007)

run-compiled 5
(1.509828065 0 0.0)
(8.29346977 1 0.3554160189999962)


run 20
(6.496330834 1 0.36043094000000053)
(32.831416548 4 1.4240319729999982)
````

`syntax-ppss` does all the relevant calling. It's worth noting that the two implementations have different complexities; mine is linear, scaling with the number of "!" till the first macro, while  the old implementation is linear with `rust-paren-level`. So the old one scales badly with deeply indented code, while mine scales badly with people who shout a lot in their comments.

I've used this sort of bench because it's more realistic; `rust-in-macro` is getting called many times in many places in a buffer.

**Comment by tspiteri** [comment]
I have [a large file](https://docs.rs/crate/rug/1.8.0/source/src/float/big.rs) which I loads quite slow in rust-mode. It feels even slower with this PR (e.g. instead of counting about 7 seconds for rust-format-buffer to finish, I count about 10 seconds with this PR; not very scientific I know, but still a good indication for me).

(Turning off Match Angle Brackets does make a very significant improvement without the PR, I haven't tested that with this PR.)

**Comment by tspiteri** [comment]
> is linear, scaling with the number of "!" till the first macro… scales badly with people who shout a lot in their comments.

I think this is what causes the regression in my file, `(count-matches "//.*!")` returns 589 because of all the assertions in the doc examples in the file.

**Comment by phillord** [comment]
@tspiteri Well, if you are going to put tests into your files, then I have no sympathy for you at all. Can't you just delete them all?

I tried this with my bench code. You obviously have a faster machine than I because it takes me 70 seconds to run `syntax-ppss`, with current code, and indeed my PR makes that 170 seconds. Both totally unusable.

Hmm. I can think of ways of speeding this up more, but not simple ones. Let me cogitate.

**Comment by tspiteri** [comment]
@phillord 
> Well, if you are going to put tests into your files, then I have no sympathy for you at all.

Ouch :)

> Can't you just delete them all?

They are doc examples, not tests. And the standard way to show the output of a function in a doc example is through an assertion, e.g. [`i32::saturating_add`](https://doc.rust-lang.org/std/primitive.i32.html#method.saturating_add). And that's what I do, e.g. the [`Float::to_i32_saturating`](https://docs.rs/rug/1.8.0/rug/struct.Float.html#method.to_i32_saturating) doc shows an example of no saturation, a saturation down, and a saturation up. Are you suggesting I should remove my examples?

> I tried this with my bench code. You obviously have a faster machine than I because it takes me 70 seconds to run `syntax-ppss`, with current code, and indeed my PR makes that 170 seconds. Both totally unusable.

That difference seems a bit excessive, did you byte compile before measuring? (I'm not really knowledgeable of the elisp environment, but this is the equivalent of the Rust "Are you using --release?", and I always have to look up how to do it when I need it.)

**Comment by phillord** [comment]
"Are you suggesting I should remove my examples?"

It was my attempt at humour. I thought that the statement was outrageous enough not be taken seriously. Apologies!

"did you byte compile before measuring?"

No, although I have given it a go now and it still takes ages. Byte-compiling doesn't normally make that much difference, especially when most of the time is spent regexp matching and moving around the buffer which is mostly likely in this case. It takes 0.24 seconds with `(setq rust-match-angle-brackets nil)`. The performance crappiness here, I believe, remains that we are operating in at least quadratic time to identify macro scope, when it should be possible with a linear scan.

Again, I will cogitate.

**Comment by tspiteri** [comment]
I tried to have a look, but the syntax stuff is all very alien to me. One idea I had is that in one “operation” there will be multiple calls to `rust-in-macro`, and they will all keep on looking back until 
either they're out of nested parens (old) or out of bangs (new). But I think they keep looking over the same text. So maybe at the start of the operation initialize a variable `bound` to 0, and then any time `rust-in-macro` returns false, set `bound` to the position. Then, whenever for example there is a call to `backward-up-list` or `rewind-irreleveant` (old) or `search-backward` (new), if the position is before `bound` we know that we don't need to keep on looking.

I don't know if this works or not, as I don't know how syntax works or what an “operation” is.

**Comment by phillord** [comment]
> I tried to have a look, but the syntax stuff is all very alien to me. One idea I had is that in one “operation” there will be multiple calls to `rust-in-macro`, and they will all keep on looking back until either they're out of nested parens (old) or out of bangs (new). 

Yep, that's exactly the problem. In your file, `rust-in-macro` is called around 900 times to syntax parse the entire file.



> But I think they keep looking over the same text. So maybe at the start of the operation initialize a variable `bound` to 0, and then any time `rust-in-macro` returns false, set `bound` to the position. Then, whenever for example there is a call to `backward-up-list` or `rewind-irreleveant` (old) or `search-backward` (new), if the position is before `bound` we know that we don't need to keep on looking.
> 
> I don't know if this works or not, as I don't know how syntax works or what an “operation” is.

The functions are a bit baroque, but it's `syntax-ppss` which is forcing the buffer parse to happen. This in turn calls `rust-syntax-propertize` which finally calls `rust-in-macro`. It would be worse if `syntax-ppss` was not cached, but the cache does get dropped regularly. I suspect that `rust-in-macro` gets called in some other places as well.

I tried something like your idea. I identify the scope of all macros in `rust-syntax-propertize`,  then `rust-in-macro` just checks whether point is in any of these scopes. Identification of scopes is done, but simply searching for `!` followed by `\w+[({` and then matching forward to the matching brace. It works, in that it brings the parse down to 4 seconds (from 70) on your very big file. Which is lovely, but for reasons I cannot fathom, it breaks indentation.

So this might be a good way forward if I can work out the breakage.

https://github.com/phillord/rust-mode/tree/fix-broken/faster-in-macro

**Comment by phillord** [comment]
This is another solution that is now nearly working. It finds the scopes in a linear scan and then uses them to answer `rust-in-macro`. It uses a safe cache through a `let` binding. There should be no problems with cache invalidation.

It needs a small extension (it does not cope with `macro-rules` which seems to have a bespoke syntax); and I will optimize further using the `start` and `end` parameters of `syntax-propretize` which should reduce it to a linear scan of the visible area. Nearly there I think.

**Comment by phillord** [comment]
So this is the performance for the latest patch. Still 6 times slower than with match-angle-brackets to nil, but a lot faster than the current implementation. Still does not handle `macro_rules` yet, which will slow it down a bit more, unfortunately.

````
;; big.rs
;; run-5
;; old, new, match-angle-brackets
;; (395.52657388100005 7 3.090328318999994)
;; (20.047772558 7 2.9489682509999966)
;; (3.7125402989999996 2 0.8323715489999941)
````

**Comment by tspiteri** [comment]
I tried this and it works great! (To test it I added
```elisp
(and (looking-at rust-re-ident)
     (goto-char (match-end 0))
     (skip-chars-forward " \t\n\r")
     (eq ?\{ (char-after)))
```
after the checks for `?\[`, `?\(` and `?\{`.)

Now `rust-in-macro` + `rust-macro-scope` are using less than `rust-is-lt-char-operator` and about the same as the `rust-in-str-or-cmnt` directly inside `rust-ordinary-lt-gt-p`.
```
- rust-syntax-propertize                            186  41%
 - rust-ordinary-lt-gt-p                            149  33%
  + rust-is-lt-char-operator                         79  17%
  + rust-in-str-or-cmnt                              45  10%
  + rust-in-macro                                    20   4%
  + rust-paren-level                                  5   1%
 + rust-macro-scope                                  29   6%
```

mookid:
HI Phil, 
thanks for that work!

I am reviewing the patch during the weekend.

rust-mode.el
```diff
@@ -267,18 +268,40 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+
+(defun rust-in-macro (&optional start end)
+  (when (> (rust-paren-level) 0)
+    (let ((scopes
+           (or rust-macro-scopes (rust-macro-scope start end))))
+      (seq-some
+       (lambda (sc)
+         (and (>= (point) (car sc))
+              (<= (point) (cadr sc))))
+       scopes))))
```
tspiteri, line 304: I don't think `seq-some` is supported in emacs 24. And anyway I think using a `dolist` with `catch`+`throw` would allow us to exit early assuming `scopes` are ordered.
```elisp
(catch 'break
  (dolist (sc the-scopes)
    (if (< (point) (car sc)) (throw 'break nil))
    (if (<= (point) (cadr sc)) (throw 'break t))))
```

tspiteri: Or the other way round, I might have assumed the wrong `scopes` order

rust-mode.el
```diff
@@ -267,18 +268,40 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+       (or start (point-min)))
+      ;; if there is a start move back to the first top level
+      (when start
+        (let ((top (syntax-ppss-toplevel-pos (syntax-ppss))))
+          (when top
+            (goto-char top))))
+      (while (search-forward "!" end t)
+        (when (and (not (rust-in-str-or-cmnt))
+                   (skip-chars-forward " \t\n\r")
+                   (or
```
mookid, line 287: nit: I usually see that as `(memq (char-after) ...)`

rust-mode.el
```diff
@@ -267,18 +268,40 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+            (setq scope (cons (list start (point)) scope)))))
+      scope)))
+
+(defun rust-in-macro (&optional start end)
+  (when (> (rust-paren-level) 0)
+    (let ((scopes
+           (or rust-macro-scopes (rust-macro-scope start end))))
+      (seq-some
+       (lambda (sc)
+         (and (>= (point) (car sc))
```
mookid, line 302: how does it compare to a binary search?

tspiteri:
I was testing some, and this reduced test case seems to confuse it:
```rust
x! { 1 < 12 }
```
I looked into it and I believe that when we first assume all `[<>]` are parens and then suppress afterwards, we are confusing things as then `(forward-list)` inside `rust-macro-scope` is confused and thinks the < matches the }. So I tried to remove `?<` and `?>` from `rust-mode-syntax-table` and then inside `rust-syntax-propertize` add the text property instead of removing it.

This fixed things for me, but then the electric pairs test failed, presumably because electric pairs look at the syntax table which now has no < and >, so it is broken. Maybe there's a better way to unconfuse the use of `forward-list` in `rust-macro-scope` without breaking the syntax table, but I don't know.

tspiteri:
I managed to get it to work by using `with-syntax-table` inside `rust-macro-scope`.

https://github.com/tspiteri/rust-mode/commits/both-opts

rust-mode.el
```diff
@@ -112,6 +112,7 @@ symbols."
   (let ((beg-of-symbol (save-excursion (forward-thing 'symbol -1) (point))))
     (looking-back rust-re-ident beg-of-symbol)))
 
+
```
tspiteri, line 115: stray

rust-mode.el
```diff
@@ -267,18 +268,40 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+        (when (and (not (rust-in-str-or-cmnt))
+                   (skip-chars-forward " \t\n\r")
+                   (or
+                    (eq ?\[ (char-after))
+                    (eq ?\( (char-after))
+                    (eq ?\{ (char-after))))
+          (let ((start (point)))
+            (forward-list)
+            (setq scope (cons (list start (point)) scope)))))
+      scope)))
```
tspiteri, line 294: Here if we are in a region without macros, scope is nil, which means that rust-in-macro will call rust-macro-scopes repeatedly. Better return `(cons scope nil)` and then in rust-in-macro use `(car scopes)` after the check for nil.

rust-mode.el
```diff
@@ -1211,34 +1234,37 @@ whichever comes first."
...
+           (unless (save-excursion (nth 8 (syntax-ppss (match-beginning 0))))
+             (put-text-property (match-beginning 1) (match-end 1)
+                                'syntax-table (string-to-syntax "|"))
+             (rust--syntax-propertize-raw-string (match-beginning 0) end)))))
+      ("[<>]"
+       (0 (ignore
+           (when (save-match-data
+                   (save-excursion
+                     (goto-char (match-beginning 0))
+                     (rust-ordinary-lt-gt-p)))
```
tspiteri, line 1263: start and end should be passed to rust-ordinary-lt-gt-p so that it passes them to rust-in-macro.

phillord:
I've updated it with a some more tests cases and stopped it breaking on unbalanced buffers. I will address the comments above next!

rust-mode.el
phillord: `scopes` is backward. The original plan I had it wasn't going to be ordered either.

If `seq-some` is supported in Emacs-24, it will be via ELPA. But, Emacs-27 is now in release-candidate. Do we really need to support Emacs-24?

rust-mode.el
phillord: Which?

rust-mode.el
phillord: Haven't timed it. Probably not that much difference because lists are linear time access. I really wouldn't be too worried about manipulation of integer lists, though, compared to all the buffer movement.

rust-mode.el
phillord: I don't understand. `rust-in-macro` should always call `rust-macro-scope`, unless `rust-macro-scopes` is `let`-bound which it is in `rust-syntax-propertize`. That's the only place `rust-in-macro` is called a repeatedly within a single turn of the command-loop.

rust-mode.el
phillord: It doesn't make any difference because `rust-macro-scopes` is `let`-bound. So `rust-in-macro` isn't using `start` and `end`.

rust-mode.el
phillord: Yep. Fixed.

rust-mode.el
tspiteri: Let's say I have a large file with not a single exclamation mark (or the region between start and end, doesn't have to be the whole file). Then rust-macro-scope sets rust-macro-scopes to nil as there are no macro scopes. Then rust-in-macro sees that rust-macro-scopes is nil so it calls rust-macro-scope every time.

rust-mode.el
tspiteri: I think I saw lots of lines like "rust-syntax-propertize 300000 305000" followed by "rust-macro-scope nil nil" when I was debugging the problem above where rust-in-macro was calling rust-macro-scope repeatedly, so that's why I mentioned this, but I think yes, fixing the other issue would make this unnecessary.

(I was going through a test run that involved editing a function near the end of big.rs, and the editing was very slow until I made these couple of changes.)

rust-mode.el
tspiteri: I don't have an opinion on supporting 24, but the github actions tests were breaking for me, so I guessed unless the maintainers decide to bump the requirement and remove that test, it needs to support 24.

https://github.com/phillord/rust-mode/actions/runs/92635709

rust-mode.el
phillord: Ah, yes, good point. Fixed this now, but in a slightly different way.

phillord:
@tspiteri With this example, 

````rust
x! { 1 < 12 }
````

Can you tell me what it is you find is confused? I can't see what is wrong.

**Comment by tspiteri** [comment]
@phillord Before the balance fixes (so on commit 05a7ef6), if I created a file with only that content, the brackets would be detected as an unmatched { and a mismatched < }. That has already been fixed by your do-not-require-balance fix.

mismatch

File: rust-mode.el (hunk 5 above)
Comment (tspiteri) [reply rust-mode.el]: It's still broken. The `defvar-local` should still be to nil as no pass has been done yet. And in rust-in-macro, scopes should still be bound to `(or rust-macro-scopes (rust-macro-scope ...))` to avoid a repeated search on `'empty`.

Comment (phillord) [reply rust-mode.el]: Oh, dear. I am missing my own "assess" library which makes testing this sort of thing easier. I believe I have fixed this properly now.

File: rust-mode.el (hunk 6 above)
Comment (phillord) [reply rust-mode.el:1287]: I have tested this and cannot reproduce, except in one circumstance. If I load the current master of rust-mode.el and then eval `in-macro` and `scopes`, so that the `let' binding in `rust-syntax-propertize' is not happening. Easy enough to get into this state actually, and I just spent 20 minutes debugging before I realised.

**Comment by phillord** [comment]
As far as I can see, we are complete now, except for one question depending on the outcome of https://github.com/rust-lang/rust-mode/pull/373 and a big squash. I am afraid I have made the code review above a bit of a mess, but are there any other needed fixes>

File: rust-mode.el (hunk 6 above)
Comment (tspiteri) [reply rust-mode.el:1287]: Oh sorry about the confusion. I only saw those repetitions *before* the other fix, with the other fix I don't see them. So this was resolved properly in the other fix and you were right, this suggestion is unnecessary.

File: rust-mode.el (hunk 2 above)
Comment (mookid) [reply rust-mode.el]: I mean to replace 
```(or  (eq ?\[ (char-after))  (eq ?\( (char-after))  (eq ?\{ (char-after)))```
 with 
```(memq (char-after) '(?\( ?\[ ?\{))```.

File: rust-mode.el (hunk 3 above)
Comment (mookid) [reply rust-mode.el]: Sure, binary search should be done on an array structure (or a binary tree kind of data structure). It could still be done for another PR, don't bother with that just yet.

File: rust-mode.el (hunk 2 above)
Comment (phillord) [reply rust-mode.el]: Oh, yes, good call. Fixed.

**Comment by phillord** [comment]
Found a bug yesterday. I'd like to use this in practice for a week (didn't write much rust in the last few days because I have been doing lisp) to see if anything else uncovers. @tspiteri I presume you are using it in real life also?

**Comment by tspiteri** [comment]
@phillord Yes I'm using it as well. In fact I had thought there was an issue with some highlighting near the point, but it turns out it was lsp-mode that was causing the issue - I have had no issues yet using rust-mode without lsp-mode, but I'll keep on using it and will let you know if anything else comes up.

**Comment by phillord** [comment]
> @phillord Yes I'm using it as well. In fact I had thought there was an issue with some highlighting near the point, but it turns out it was lsp-mode that was causing the issue - I have had no issues yet using rust-mode without lsp-mode, but I'll keep on using it and will let you know if anything else comes up.

That might also have been the bug from yesterday; errors during syntax highlighting are demoted, and only show as messages, but leave the highlighting broken.

**Comment by tspiteri** [comment]
It could be: I can't reproduce it today while two days ago I could easily. What threw me off is that when I disabled lsp-mode just to make sure, for some reason I wasn't be able to reproduce it.

On another note, I found an issue with nested macros:
```rust
macro_rules! outer {
    () => { vec![] };
}
// rust-macro-scope returns ((46 48) (20 31))
// which correspond to ("[]" "{\n    () =>")
```

tspiteri:
A false positive for rust-macro-scope occurs with the bitwise complement of a bracketed expression.
```rust
fn foo<T>() {
    if !(mem::size_of::<T>() > 8) {
        bar()
    }
}
// rust-macro-scope returns ((23, 45))
// which corresponds to ("(mem::size_of::<T>() >")
```

tspiteri:
And in the previous test case above, there was also a mismatch with the "> 8" bit but it did not hurt bracket matching. However below there is a mismatch which does break bracket matching. (This time the exclamation mark is in a proper macro to keep the issues separate.)
```rust
fn foo<T>() {
    if mac!(mem::size_of::<T>() < 8) {
        bar()
    }
}
// rust-macro-scope returns ((26, 75))
// which corresponds to ("(mem ... bar()\n    }\n}")
```

rust-mode.el
```diff
@@ -260,18 +260,88 @@ to the function arguments.  When nil, `->' will be indented one level."
         ;; Rewind until the point no longer moves
         (setq continue (/= starting (point)))))))
 
-(defun rust-in-macro ()
+(defvar-local rust-macro-scopes nil
+  "Cache for the scopes calculated by `rust-macro-scope'.
+
+This variable can be `let' bound directly or indirectly around
+`rust-macro-scope' as an optimization but shoult not be otherwise
```
mookid, line 267: minor: typo

rust-mode.el
```diff
@@ -260,18 +260,88 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+set.")
+
+(defun rust-macro-scope (&optional start end)
```
mookid, line 270: if all call sites use the arguments, why not make them mandatory?

rust-mode.el
```diff
@@ -260,18 +260,88 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+the caller is responsible for ensuring that the data in
+`rust-macro-scopes' is up to date."
+  (when (> (rust-paren-level) 0)
+    (let ((scopes
+           (or
+            rust-macro-scopes
+            (rust-macro-scope start end))))
+      ;; `rust-macro-scope' can return the symbol `empty' if the
+      ;; buffer has no macros at all.
+      (when (listp scopes)
```
mookid, line 339: minor: this line could be `(when scopes` and remove the `'empty` thing, no?

mookid: oh ok I get why you want a non-nil thing. please ignore that comment.

rust-mode.el
```diff
@@ -260,18 +260,88 @@ to the function arguments.  When nil, `->' will be indented one level."
...
+              (when (re-search-forward "[[({]" nil t)
+                (backward-char)
+                (let ((start (point)))
+                  (ignore-errors (forward-list))
+                  (setq scope (cons (list start (point)) scope)))))))))
+      ;; Return 'empty rather than nil, to indicate a buffer with no
+      ;; macros at all.
+      (or scope 'empty))))
+
+(defun rust-in-macro (&optional start end)
```
mookid, line 322: capturing that rust-in-macro caching work in a unit test would be nice, I believe it is likely to get broken for some reason.

rust-mode.el
phillord: Fixed

rust-mode.el
phillord: Yes. Actually, they were effectively mandatory because I used `end` without a nil check.

rust-mode.el
phillord: I can test to see whether we can break caching, or fake caching. Checking whether it actually ever caches is a bit hard -- I'd have to call-capture calls to `rust-macro-scope'. I'd generally use my own library for this https://github.com/phillord/assess/blob/master/assess-call.el.

But that would add a dev dependency to rust-mode.el which adds it's own layer of complexity.

mookid: I was thinking about a cheap version of that: just advice the function that really does the work to count the number of calls, and check that it's 1.

phillord:
@tspiteri Well, that's hard. It depends on how rigid and accurate we want to be. Consider this:

````rust
fn foo<T>() {
    if !(mem::size_of::<T>() > 8) {
        bar()
    }

    dbg !(mem::size_of::<T>() > 8) {
        bar()
    }

    dbg!{hello}
}
````

I can distinguish between the `if` and last `dbg!{}` straight-forwardly using `rust-looking-back-macro`. But between the first `dbg` with a space (`dbg !`), well that's harder. I guess it rust achieves this because `if` is a keyword.

I'm a bit loath to add that sort of complexity into the checking, so my suggestion would be to fix this but accept that it breaks the second usage `dbg !()`. WDYT?

tspiteri:
I agree with your wariness to complexity, and I like your solution. After all `dbg !()` is quite uncommon and I think considered bad style. Handling it would not be worth any loss of performance and potential bugs from the extra complexity.

phillord:
I'll do that they. I actually had to check whether `dbg !()` was valid syntax when I wrote the first version of it.

phillord:
@tspiteri 

````rust
macro_rules! outer {
    () => { vec![] };
}
;; returns ((38 40) (20 45)))
````

I think this is working as I expected. So, something appears to have fixed this in the last few days.

Been a long haul, this fix, but nearly there, I hope!

tspiteri:
I opened the files I had left for those cases and bracket matching works as expected in them.

I've been using this and basically now I can leave Match Angle Brackets On with the same feel as with it off, so for me it's already there. And I wouldn't worry about the corner cases: the worse it would probably do is maybe break some bracket matching around some weird macros, but even without this patch, I used to get mismatched brackets all the time when I used Angle Brackets On; in fact I get the feeling that even bracket matching has improved, not just performance.

phillord:
I have written some code now with this and @tspiteri  hasn't found anything else, so I think we are good to go here. If we find any other problems, I can fix them on master.

mookid:
> I can fix them on master.

Indeed.

If you prefer to preserve some history feel free to push again.

Thanks @phillord and @tspiteri for the work on this!

tspiteri:
Thanks @phillord!

Issue #350, opened 2020-01-28
rust-format-buffer resets all code blocks to "shown" under hideshow


codesections:
Thanks for the great package.

I use `rust-mode` along with the builtin Emacs `hs-minor-mode`.  Unfortunately, the two do not play very well together.  Specifically, whenever I run `rust-format-buffer` and rustfmt makes any changes to the code, all code is displayed again – even code that was previously marked as hidden using hideshow.  If rustfmt does *not* make any changes, then no hidden code is incorrectly shown.

I suspect that this is related to #260 (and #143 mentions overall integration with HideShow).

mookid:
hi @codesections, I can't reproduce any more, likely since #371. Closing it, feel free to reopen if it is still a problem.

Issue #260, opened 2018-01-15
rust-format-buffer breaks markers in buffer


tspiteri:
For example

1. Open this file *a.rs*:
   ```rust
   // some function
   fn foo() {        }
   // some other function
   fn bar() {}
   ```
2. `M-x grep` and search for `fn` in *a.rs*.
3. `M-x next-error` (or `M-g n`), which moves the mark to line 2.
4. `M-x rust-format-buffer`, which removes the excess spaces between the braces in line 2.
5. `M-x next-error` moves the mark to the beginning of the buffer instead of line 4.

As a workaround, I can insert step 4.5: save buffer, and reopen it with `M-x find-alternate-file` (or `C-x C-v`), after which the broken markers for the buffer are lost and the next `M-x next-error` will move the mark to line 4. This is not perfect, as any edits that change the line numbers of matches will now cause an offset, but it is much better than having all errors sending me to the beginning of the buffer.

So if there is a way to invalidate all the markers in the buffer after a `M-x rust-format-buffer` which changes the buffer, that would basically solve this issue.

Dushistov:
May be relate to #256

nikomatsakis:
I've noticed this too, it's annoying =)

mookid:
@tspiteri I can't reproduce any more, can we close the issue?

tspiteri:
Yes, it should be fixed on emacs >= 26.2. For me that is enough: I have no opinion on whether the issue should be kept open because of emacs 25 and 26.1.

PR #371, opened 2020-04-26
In emacs >= 26.2, use replace-buffer-contents after formatting

tspiteri:
This saves the position and the markers in the buffer, so there is no need to save positions of direct and indirect buffers and windows. This also plays better with e.g. grep, as otherwise all grep markers would be destroyed when replacing the contents of the buffer with copy-to-buffer.

While `replace-buffer-contents` was added in Emacs 26.1, it was broken for non-ASCII contents and would corrupt the buffer contents. This was fixed in 26.2, so we require version 26.2 for this instead of just relying on `fboundp` (which still has to be used for byte compilation on Emacs < 26).

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1463,7 +1463,13 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
              ((zerop ret)
               (if (not (string= (buffer-string)
                                 (with-current-buffer buf (buffer-string))))
-                  (copy-to-buffer buf (point-min) (point-max)))
+                  ;; replace-buffer-contents was in emacs 26.1, but it
+                  ;; was broken for non-ASCII strings, so we need 26.2.
+                  (if (and (fboundp 'replace-buffer-contents)
+                           (version<= "26.2" emacs-version))
+                      (with-current-buffer buf
+                        (replace-buffer-contents rust-rustfmt-buffername))
+                    (copy-to-buffer buf (point-min) (point-max))))
               (kill-buffer))
              ((= ret 3)
               (if (not (string= (buffer-string)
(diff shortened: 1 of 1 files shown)


tspiteri:
Fixes #260.
Fixed #350.
(Only for emacs version >= 26.2, but I think that's reasonable.)

mookid:
Thanks for the investigation!

Alternatively to that conditional code, how costly (in terms of lines of code) would it be to backport the fixes made in upstream emacs to make that work?  That would avoid the conditionals everywhere.

tspiteri:
The `replace-buffer-contents` function was new in 26.1, and it is in C not in elisp, so I don't think it can be backported. And the essential 26.2 fix is also of course in the C code and cannot be backported. Porting the functions to elisp would be very costly I think; both in the sense that would be a major undertaking to develop, and in the sense that it would have a large performance drop in a function that already can be very heavy if the formatting does a lot of changes.

mookid:
> The `replace-buffer-contents` function was new in 26.1, and it is in C not in elisp,

Ok, I did not think about the case of a fix in the C code.
Thanks for the info.

rust-mode.el
```diff
@@ -1463,7 +1463,13 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
              ((zerop ret)
               (if (not (string= (buffer-string)
                                 (with-current-buffer buf (buffer-string))))
-                  (copy-to-buffer buf (point-min) (point-max)))
+                  ;; replace-buffer-contents was in emacs 26.1, but it
+                  ;; was broken for non-ASCII strings, so we need 26.2.
+                  (if (and (fboundp 'replace-buffer-contents)
+                           (version<= "26.2" emacs-version))
```
mookid, line 1469: If I understand well, `(version<= "26.2" emacs-version)` should be enough, right?

tspiteri: That's what I thought too :) But then the tests fail for emacs 24 and 25 because there is a call to an unbound function. 

https://github.com/tspiteri/rust-mode/actions/runs/88559854

```
Run ./run_rust_emacs_tests.sh
Byte-compilation failed:

In end of data:
rust-mode.el:2108:1:Warning: the function `replace-buffer-contents' is not
    known to be defined.
##[error]Process completed with exit code 4.
```

mookid: Ok, if you don't know hacks to calm the byte compiler let's leave it as is.

PR #372, opened 2020-04-28
Check for -> and => early in rust-ordinary-lt-gt-p

tspiteri:
Since the check for `->` and `=>` is very cheap, move it up in `rust-ordinary-lt-gt-p` potentially saving much more expensive checks.

Also use simple equality check instead of regex function `looking-at` for checking following character.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1139,6 +1139,10 @@ should be considered a paired angle bracket."
    ;; If matching is turned off suppress all of them
    ((not rust-match-angle-brackets) t)
 
+   ;; This is a cheap check so we do it early.
+   ;; Don't treat the > in -> or => as an angle bracket
+   ((and (= (following-char) ?>) (memq (preceding-char) '(?- ?=))) t)
+
    ;; We don't take < or > in strings or comments to be angle brackets
    ((rust-in-str-or-cmnt) t)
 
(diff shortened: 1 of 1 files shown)


tspiteri:
While @phillord is working on optimizing `rust-in-macro` in #369 this PR improves `rust-ordinary-lt-gt-p` to reduce the number of times `rust-in-macro` is called.

mookid:
Hi @tspiteri,

as for the other PR, some metrics with an example file (like the one you mentionned on the other thread) are very welcome (and would be useful for investigations in the future as well).

The patch looks great (I like the small efficiency gains looking-at => following-char / preceding-char and member => memq). I'd like to take some time to test it and maybe add some test in the suite to capture the behaviour of that code, but I am all for merging.

tspiteri:
With the same [large file](https://docs.rs/crate/rug/1.8.0/source/src/float/big.rs) I mentioned in the other PR, the number of calls to `rust-in-macro` fell from 935 to 562, which is a fall of about 40%.

Benchmarking using  the same code:
```elisp
(defun bench ()
  (goto-char (point-max))
  (message "%s"
           (benchmark-run 50
             (syntax-ppss-flush-cache 1)
             (syntax-ppss))))
```

Before: `(224.784685687 3 0.17687254000000002)`
After: `(114.610272863 3 0.205725889)`

tspiteri:
I also benchmarked big.rs with/without this patch after applying the latest version of the in-macro-faster patch. While the effect of this patch is reduced, as now rust-in-macro is way faster, this patch still saves an extra 10% in time.

Only in-macro-faster: `(9.936009942999998 2 0.09233385400000002)`
Both in-macro-faster and this: `(8.882999213 2 0.10287831900000002)`

Summary bench results:

|  | Without this | With this |
|---|---|---|
| Without in-macro-faster | 224.8 | 114.6 |
| With in-macro-faster | 9.9 | 8.9|

PR #373, opened 2020-05-02
Drop Emacs-24

phillord:
Emacs-24 is now two major versions behind and will soon be three, so
formally remove support for it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,6 @@ jobs:
       matrix:
         rust_version: [stable]
         emacs_version:
-          - 24.5
           - 25.1
           - 25.2
           - 25.3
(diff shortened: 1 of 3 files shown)


mookid:
I think that if support is easy to maintain, just do that. If that's a matter of rewriting a usage of seq-some, I believe we should just do that.

phillord:
Emacs-24 is becoming harder and harder to support. Yes, it is the use of `seq-some` which bought this to mind, but it is not just that. In fact, both the docker nor github actions environment test only Emacs 24.5, although there were significant changes (to advice for instance) between 24.3 and 24.4 which should probably have resulted in a major version bump. So, rust-mode probably doesn't really support Emacs-24, just some of it.

It would be nice to be able to use seq.el properly, but also as you can see from this patch also `defvar-local` and `up-list` backward compatibility hacks can be removed. So it would nice to remove an Emacs-24 requirement.

But, we are about to move to Emacs-27 at which point rust-mode will be supporting three major versions (or looked at in another way, ignoring 8 years of improvements to core). At this point I believe dropping Emacs-24 becomes somewhat of a moral obligation!

Anyway, that's all of my arguments! Happy to move forward with what ever you decide.

brotzeit:
I agree with @phillord. emacs24 is almost 8 years and most contributors are using newer versions and don't want to make their lives harder when improving rust-mode. @mookid it's your decision, but we have to drop it at some point...

mookid:
Ok, let's try it. If there are many user complains we can still revert that decision.
Thanks @phillord @brotzeit !

phillord:
Yes, happy to revert if we get complaints; a good compromise indeed.

PR #374, opened 2020-05-08
Change 'rust-build' to 'rust-compile' in README.md

nlopes:
The right name is `rust-compile` not `rust-build` so this should make it clear for new users.


mookid:
Thanks.

PR #375, opened 2020-05-09
Use <kbd> to indicate keyboard input in README

chwarr:
Fixes the flow of the sentence "Commands like TAB should indent
correctly." Previously, on GitHub, this read as "Commands like should
indent correctly."

On GitHub, `<kbd>` elements also look like keyboard keys.


mookid:
thanks!

PR #376, opened 2020-05-17
Fix various byte compilation warnings in rust-mode-tests.el.

phst:
- Migrate to lexical binding.
- Declare functions that are conditionally-defined.
- Migrate off the deprecated ‘cl’ library.
- Don’t use ‘setq’ to introduce global variables and constants.
- Avoid “unused lexical variable” warnings.
- Remove some unused variables.
- Stop using interactive-only commands.


mookid:
Thanks.

PR #380, opened 2020-05-21
Warn the user if rustfmt does not support the --edition flag.

### LINKED ISSUE #361: Detect old versions of rustfmt.
Currently the `rust--format-call` function calls `rustfmt` using by default the flags `--edition 2018`. It fails badly if the version of the binary is old enough, in a very unhelpful way.

rust-mode should output a helpful error message in the echo area in case that happens. See #360.

mookid:
fix #361


mookid:
@brotzeit  what do you think? maybe it is a bit overkill.

brotzeit:
I would just display the output of rustfmt.

mookid:
Sounds fair. This is what is currently done.

PR #393, opened 2020-07-06
Make more easily comparable with rustic fork

tarsius:
A few months ago it looked like I was going to learn Rust and so I looked for the best Emacs major-mode.  (Actually learning Rust got delayed, but I still plan on doing it.) I found this `rust-mode` and the `rustic` fork by @brotzeit.

I think that fork was a bit unfortunate.  What @brotzeit mostly did was to add extra functionality but he also refactored the existing essential major-mode/syntax/... parts.  Many of these changes were rather superficial, but because I did not backport them immediately *and* made major rearrangements it became very hard to compare the respective bits from the two implementations.

These commits, which I wrote five months ago for the most part, try to fix that.  I already got a lot of cleanup merged into `rustic` before, what these commits here to is to arrange `rust-mode.el` the same way `rustic.el` has been rearranged before. Other commits port doc-string improvements, other cleanup and the addition of `outline-minor-mode`-compatible section headings.

After applying these changes you can copy `rustic.el` from the other repository, rename it to `rust-mode.el` and also do `s/rustic/rust` inside and then you get pretty useful diff. (I do not think that *all* rearrangement is an improvement or even just necessary. Some are, but others don't really make a difference either way.)

My hope is that I can get you guys to cooperate some more, so that eventually there won't be two major-modes anymore. I think it makes sense for there two be two projects, a battery-included thing like `rustic` and a just-the-major-mode `rust-mode`, but it would be great if `rust` could just use the `rust-mode.el` from `rust-mode`.

I have written these commits and a lot of cleanup to `rustic` that has already been merged, but I think that from here you will have to pick it up and make the fork a thing of the past.

Thanks for considering this proposal,
and lots of fun with the commits!

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -137,7 +137,7 @@ to the function arguments.  When nil, `->' will be indented one level."
   "Face for interpolating braces in builtin formatting macro strings."
   :group 'rust-mode)
 
-;;; Rust-mode
+;;; Syntax
 
 (defun rust-re-word (inner) (concat "\\<" inner "\\>"))
 (defun rust-re-grab (inner) (concat "\\(" inner "\\)"))
@@ -229,6 +229,8 @@ Use idomenu (imenu with `ido-mode') for best mileage.")
 
     table)
   "Syntax definitions and helpers.")
+
+;;; Mode
 
 (defvar rust-mode-map
   (let ((map (make-sparse-keymap)))
@@ -934,7 +936,8 @@ buffer."
   (save-excursion (= (progn (goto-char pos1) (line-end-position))
                      (progn (goto-char pos2) (line-end-position)))))
 
-;; Font-locking definitions and helpers
+;;; Font-locking definitions and helpers
+
 (defun rust-next-string-interpolation (limit)
   "Search forward from point for next Rust interpolation marker before LIMIT.
 Set point to the end of the occurrence found, and return match beginning
@@ -1372,6 +1375,8 @@ whichever comes first."
         (put-text-property (1- (match-end 2)) (match-end 2)
                            'syntax-table (string-to-syntax "|"))
         (goto-char (match-end 0))))))
+
+;;; Syntax Propertize
 
 (defun rust-syntax-propertize (start end)
   "A `syntax-propertize-function' to apply properties from START to END."
@@ -1911,6 +1916,8 @@ Return the created process."
   (interactive)
   (compile (format "%s test" rust-cargo-bin)))
 
+;;; Hooks
+
 (defun rust-before-save-hook ()
   (when rust-format-on-save
     (condition-case e
@@ -1928,6 +1935,8 @@ Return the created process."
         ;; would show "Wrote ..." instead of the error description.
         (or (rust--format-error-handler)
             (message "rustfmt detected problems, see *rustfmt* for more."))))))
+
+;;; Compilation
 
 (defvar rustc-compilation-location
   (let ((file "\\([^\n]+\\)")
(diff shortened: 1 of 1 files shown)


mookid:
Some tests fail. I suggest to make sure that tests pass at each step of the refactoring, as it is easy to omit some code.

rust-mode.el
```diff
@@ -782,7 +785,7 @@ Returns nil if the point is not within a Rust string."
     "println")
   "List of builtin Rust macros for string formatting.
 This is used by `rust-font-lock-keywords'.
-(`write!' is handled separately).")
+\(`write!' is handled separately).")
```
mookid: is that necessary? I can't see a difference in the generated doc.

rust-mode.el
```diff
@@ -137,25 +137,9 @@ to the function arguments.  When nil, `->' will be indented one level."
   "Face for interpolating braces in builtin formatting macro strings."
   :group 'rust-mode)
 
-(defun rust-re-word (inner) (concat "\\<" inner "\\>"))
-(defun rust-re-grab (inner) (concat "\\(" inner "\\)"))
-(defun rust-re-shy (inner) (concat "\\(?:" inner "\\)"))
```
mookid, line 33: at least some of these have to be before the vars definitions.

rust-mode.el
```diff
@@ -2121,6 +2115,12 @@ visit the new file."
                (delete-pair))
               (t (rust-insert-dbg)))))))
```
mookid, line 2133: I think that this can be removed.

mookid:
I made a few comments concerning a couple details (I reviewed commit by commit). I am ok to merge when tests pass at each step of the refactoring, and I believe that this goes in the right direction.

However, I don't really know what can be achieved on the mid term; I miss time to work on this, and I prefer to spend that time improving the code here rather than hypothetically converging with the fork. And not many other people want to work on rust-mode.

tarsius:
Great! I must admit I did not test after rebasing this a few times on top of new changes. But knowing that you intend to merge, the motivation is back and I will give this the necessary final polish. Stay tuned.

tarsius:
Tests pass now. The only problem I found was those few functions that were placed after they are used while setting the initial values of some variables; which you pointed out inline.

rust-mode.el
tarsius: Fixed.

rust-mode.el
tarsius: Maybe, I don't know.
I don't think my commits touch this.
This should be addressed separately.

tarsius:
> However, I don't really know what can be achieved on the mid term; I miss time to work on this, and I prefer to spend that time improving the code here rather than hypothetically converging with the fork. And not many other people want to work on rust-mode.

No problem. I wanted to land this now because dealing because resolving conflicts when rebasing onto incoming changes became a bit annoying. I probably won't contribute anything *soon*, but eventually I probably will return. ;D

brotzeit:
@tarsius thanks for your work. I'm absolutely in favor of reducing the maintenance effort for both packages. However I'm still not sure how we want to treat similiar functions that exist in both packages. For example `rust-format-buffer` in rust-mode works differently than the equivalent in rustic. Do you have any idea how we can solve this ?

mookid:
Looks good to me, merging tomorrow if nobody screams. Thanks for the work!

tarsius:
@all Cheers!

@brotzeit Those can coexist. Maybe you have to add some safety wheels to prevent that both of them are used at them same time, but I don't see a problem here.

brotzeit:
Hmmm. Ok I will try to do it incrementelly somehow.

PR #395, opened 2020-07-09
Use proper compilation mode for *rustfmt* buffer.

### LINKED ISSUE #204: rustfmt should display errors

mookid:
A proposal to fix #204.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1600,7 +1600,8 @@ This is written mainly to be used as `end-of-defun-function' for Rust."
 (defun rust--format-call (buf)
   "Format BUF using rustfmt."
   (with-current-buffer (get-buffer-create rust-rustfmt-buffername)
-    (view-mode +1)
+    (compilation-mode 1)
+    (setq-local compile-command (format "%s %s" rust-rustfmt-bin (buffer-file-name buf)))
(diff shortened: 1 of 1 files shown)


brotzeit:
Maybe we should use `define-compilation-mode` and give the mode a different name to avoid confusion.

PR #397, opened 2020-08-11
compilation: interpret ANSI color code.

### LINKED ISSUE #184: Colored error messages
Emacs should display error messages using the same colors emitted by the compiler.

mookid:
fix #184.
with and without the new code:

image


brotzeit:
Sorry for not giving an approval though I looked at it. I'm a little hesitant since I don't know for sure if this will work. But if you didn't encounter any issues while using it, it's probably fine ?

birkenfeld:
While this works for the simple functions `rust-compile` etc., it does not for generic `compile`. Hence there is no way to customize the command line used; for example I often need to either `cd subdir; cargo build` or `cargo run --bin something`.

I've hacked together an alternative at #401.

PR #399, opened 2020-11-21
Add let-binding for json-object-type around json-read call 

Lispython:
Current version of rust-buffer-project can fail if somewhere json-object-type settled to hash-table.

This MR add let*-binding for use json-object-type with alist type.


jimblandy (approved):
Thank you!

PR #400, opened 2020-12-04
Add rust-check function

alyssais:
Because check exists and is so much faster, I only ever want to build
when I want to test the binary, in which case I'd use run.  So I think
it's more useful to have a rust-check function available than it is to
even have rust-compile.


mookid:
thanks!

PR #401, opened 2020-12-06
compilation: force colored cargo output

birkenfeld:
Alternative to #397 which works for any invocation of `compile`.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -16,7 +16,8 @@
 
 (eval-when-compile (require 'rx)
                    (require 'compile)
-                   (require 'url-vars))
+                   (require 'url-vars)
+                   (require 'ansi-color))
 
 (require 'json)
 (require 'thingatpt)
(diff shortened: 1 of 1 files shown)


brotzeit:
If somebody else can test this I would be happy to merge this PR.

brotzeit:
@mookid what do you think ?

rust-mode.el
```diff
@@ -16,7 +16,8 @@
 
 (eval-when-compile (require 'rx)
                    (require 'compile)
-                   (require 'url-vars))
+                   (require 'url-vars)
+                   (require 'ansi-color))
```
nbfalcon, line 20: `ansi-color-apply-on-region` is a function, not a macro, so shouldn't this *not* be in `eval-when-compile`?

psibi:
I'm gently closing this PR for now as there haven't been any recent updates. However, feel free to reopen it whenever you're ready to pick it up again.

Issue #405, opened 2021-04-15
Towards making rustic use rust-mode


tarsius:
`rustic` was created as a fork of `rust-mode`, and as a consequence the former contains a fork of the Rust major-mode itself.

`rustic` takes a batteries-included approach, while `rust-mode` for the most part sticks to just the major-mode parts (basically font-lock and some movement commands).

IMO there is value in both approaches and it is nice that users can pick a package according to their preference.  But this schism also causes some problems.

1. `rustic` comes with a fork of the major-mode from `rust-mode` and changes to the former have not been back-ported.  Many new features have been added to `rustic` without there being a clear separation between the major-mode and the extra-functionality parts, making it increasingly hard to do the back-porting.  This is only getting worse over time.
   
   I have been tackling this issue by moving the non-major-mode parts out of both `rust-mode.el` and `rustic.el`, and performing other rearrangements, so that the two variants of the major-mode can be once more be compared and even be merged.
   
2. `rust-mode` also contains some functionality that goes beyond just the major-mode parts, but the implementations are rather rudimentary.  These features have been improved in `rustic` and it might be worth thinking about completely removing some of them from `rust-mode`.
   
   This is related to (1) and the cleanup I have done to tackle that will help here as well, but this does not have to happen at the same time.
   
At this time I am opening two pull-request to address (1).

- TODO add cross-reference (rustic)
- TODO add cross-reference (rust-mode)

On top of the latter I have created a few more commits that split the remaining differences into easily digestible chunks.  You can see the result here:

- TODO add cross-reference (rust-mode diffs)
   
Most of the commits are prefixed with `[IGNORE FOR NOW]` but you might still want to take a look, if only to get an idea of what the remaining differences are.

The last commit consists of actual differences in how the two major-modes do font-locking.  I think these changes from `rustic` should be backported to `rust-mode.el` and I think @brotzeit as their primary author should do it.  I have no opinion on whether that should be done using one big "Backport font-lock changes from rustic" commit or the changes should be backported one-by-one.
  
After that, some more differences have to be taken care of and I can do most of that work.  Once that is done we can remove the major-mode parts from `rustic.el` and essentially replace them with

```elisp
(require 'rust-mode)
```

I first brought this up in https://github.com/brotzeit/rustic/issues/92.

brotzeit:
I think you know that I'm probably the only one replying here ;)

Give me some time to take a look at the pull requests at both repos. BTW if you want write access for rust-mode/rustic just say something.

[...]

PR #406, opened 2021-04-15
Separate the major-mode parts from the additional functionality

tarsius:
The goal is to eventually allow `rustic` to use `rust-mode.el` without also having load other parts.  Some of these commits just make it simpler to compare the two Rust major-mode implementations.

See https://github.com/rust-lang/rust-mode/issues/405 for more information and the sibling pull-request https://github.com/brotzeit/rustic/pull/246.


brotzeit:
Merging now. We'll fix the tests afterwards.

PR #409, opened 2021-06-28
Improve rust--format-fix-rustfmt-buffer

sergv:
I noticed that two loops can be merged into one. Also `save-match-data` will make this function not clobber global match data (not that anyone was suffering from that but nonetheless).

Upd: also noticed that `*rustfmt*` name can be configured but in error messages it'll stay the same. Fixed that too.


brotzeit:
LGTM

PR #424, opened 2021-11-22
locate workspace rather than member crate

rich-murphey:
This fixes a failure in `next-error`  after compiling a member of a workspace.

When invoking `rust-check` from rust source code file that is a member of a workspace, `next-error` fails because the compilation buffer lists source code paths relative to the workspace root, not the member crate.

rust-cargo.el uses cargo-locate to determine the root of the path to the source files, and this provides the location of the member crate rather than the workspace.  The compilation buffer which contains paths relative to the workspace root, not the member project.

This patch searches for a Cargo.toml in the parent and the parent's parent directories, and if present, uses that directory as the workspace root.

### DIFF
--- a/rust-cargo.el
+++ b/rust-cargo.el
@@ -1,3 +1,6 @@
+;; (byte-recompile-directory "~/gnu/rust-mode/" 0 t)
+;; (byte-compile-file "rust-cargo.el" nil)
+
 ;;; rust-cargo.el --- Support for cargo                -*-lexical-binding: t-*-
 ;;; Commentary:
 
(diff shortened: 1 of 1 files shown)


rich-murphey:
After further testing, this has an additional issue.  With the above patch, `cargo check` will run in the workspace root rather than the member crate.    So, `next-error` works, but this causes the whole workspace to be compiled.

rich-murphey:
I'm closing this because rustic has a feature that seems to cover this use case.  `rustic-buffer-workspace` seems to support compiling in workspaces.

PR #426, opened 2021-12-12
Highlight funcalls

gnuvince:
One feature I quite like of Emacs's [go-mode](https://github.com/dominikh/go-mode.el) is how it highlights function calls. I use a theme that has very few colors, and highlighting the calls helps to guide my eyes. I replicated the feature for rust-mode. Here's a screenshot.

image

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -389,7 +389,7 @@ r\"foo\\\", \"bar\", r\"bar\";
 r\"foo\\.\", \"bar\", r\"bar\";
 r\"foo\\..\", \"bar\", r\"foo\\..\\bar\";
 r\"\\\", \"foo\", r\"\\foo\";
-not_a_string();
+not_a_string;
 
 "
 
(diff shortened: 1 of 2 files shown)


brotzeit:
Thanks for the contribution! But it seems to me that this isn't the default behavior. Look at other language major modes like elisp. Function calls have the face `default`.

But we could add this as optional feature.

brotzeit:
Any opinions on this ? If users like this feature we can merge this anyway...

brotzeit:
Does this work with treesitter ? I guess font-locking will change a lot with this new feature.

psibi:
@gnuvince Now that treesiter mode support is merged, can you see if it works with that: https://github.com/rust-lang/rust-mode?tab=readme-ov-file#tree-sitter ?

psibi:
I'm gently closing this PR for now as there haven't been any recent updates. However, feel free to reopen it whenever you're ready to pick it up again.

PR #427, opened 2021-12-16
Fix `rust-electric-pair-skip-self-wrap`

roysc:
nil is a valid value for `electric-pair-skip-self`, so handle that case before trying to call it as a function


brotzeit:
Thanks!

PR #428, opened 2021-12-24
Add a face for the ampersand reference mark

Chris00:
Customizing this face may improve code readability for some users.
By default nothing is set for the face for backward compatibility.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -101,6 +101,11 @@ to the function arguments.  When nil, `->' will be indented one level."
   "Face for the question mark operator."
   :group 'rust-mode)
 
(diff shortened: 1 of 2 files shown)


rust-mode.el
```diff
@@ -101,6 +101,11 @@ to the function arguments.  When nil, `->' will be indented one level."
   "Face for the question mark operator."
   :group 'rust-mode)
 
+(defface rust-ampersand-face
+  '()
```
brotzeit, line 105: I think it's common practice to use `'((t :inherit default))`.

Chris00: I've followed your suggestion.

PR #429, opened 2021-12-24
Enable the use of prettify-symbols-mode

Chris00:
Make `M-x prettify-symbols-mode` prettify some symbols (e.g. `<=`, `>=`, `->`) in Rust code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,21 @@ on save:
 (setq rust-format-on-save t)
 ```
 
+### Prettifying
+
+You can toggle prettification of your code by running `M-x
+prettify-symbols-mode`.  If you'd like to automatically enable this
+for all rust files, add the following to your init.el.
+
+```elisp
+(add-hook 'rust-mode-hook
+          (lambda () (prettify-symbols-mode)))
+```
+
+You can add your own prettifications to `rust-prettify-symbols-alist`.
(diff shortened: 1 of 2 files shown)


Chris00:
The code to disambiguate the two uses of `||` has to be improved but, before, I'd like to know whether there is interest in thos feature.

brotzeit:
> I'd like to know whether there is interest in this feature.

I'm sure many users want this feature. But I think these changes should be optional. I'm actually not sure if we should make this a default. We should definitely document this in the readme. Most use doom-emacs etc. anyway so there will be probably a pull request for those distros soon.

I wonder if there's already some mode that provides this feature major mode independent. I think we should ensure this feature doesn't interfere with other packages.

Chris00:
>  But I think these changes should be optional. I'm actually not sure if we should make this a default.

The commit prepares the infrastructure for prettifying but nothing gets activated until the user issues `M-x prettify-symbols-mode` (or puts it in the mode hook or uses `global-prettify-symbols-mode`).  I've changed a comment to reflect that.  Prettify _is_ the major-mode independent way of prettifying symbols (but it must be tailored to the mode).

brotzeit:
Sorry, I was obviously too lazy to google. Can you do me a favor and a note to the readme ?

Chris00:
Done.

brotzeit:
Thanks!

PR #431, opened 2021-12-28
Avoid obsolete calling convention of ‘looking-back’.

phst:
Calling ‘looking-back’ with only one argument causes a compile warning.


brotzeit:
Thanks!

PR #436, opened 2022-01-18
rust-mode should be enabled automatically when opening .rs files

daniel-wong-dfinity-org:


brotzeit:
This shouldn't be necessary since it's already set https://github.com/rust-lang/rust-mode/blob/master/rust-mode.el#L312.

Maybe there's something wrong with your config.

daniel-wong-dfinity-org:
lol I dunno. All I know is that I followed the instructions, but rust-mode would not activate automatically, unless I added this to my emacs config.

My emacs version: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109))
 of 2019-09-02

Should I turn this into an issue?

tromey:
> This shouldn't be necessary since it's already set https://github.com/rust-lang/rust-mode/blob/master/rust-mode.el#L312.
> 
> Maybe there's something wrong with your config.

The issue is that the instructions for manual installation say to use `autoload`, but of course that won't cause the `add-to-list` to be evalled.  So I tend to think this patch is ok, since it updates the instructions to reflect a reasonable manual installation approach.

brotzeit:
Oh, you're right. Didn't think of it.

Thanks for the contribution!

PR #444, opened 2022-04-27
Use Eask for CI

jcs090218:
This patch does the following:

#### In `.github/workflow/test.yml`

* Use `jcs090218/setup-emacs` instead of `purcell/setup-emacs`
  - Add retry capability to prevent failure on Emacs' installation
  - Support windows
* Add setup-node (`16`)
* Remove `setup-cask` (replace with `setup-eask`)
* Add test for Emacs `28.1`
* Add tests on `Windows` and `macOS`

#### In `Makefile`

* Replace Emacs command-line arguments `emacs --batch ...` with just `eask <command>`
* Remove `ELS` variable, manage with `Eask`-file (use command `eask files` to list all package files)
* Add `make checkdoc` and `make lint` but currently disabled, see error log [here](https://github.com/jcs-PR/rust-mode/actions/runs/2235278555).
* Add `make build` to test packaging and installing.
* Replace `make xxx-autoloads` with just `eask autoloads`

#### Others

* `Eask`-file added

---

This is totally optional, I think this would make CI a lot easier.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,27 +4,33 @@ on: [push, pull_request]
 
 jobs:
     unix-test:
-        runs-on: ubuntu-latest
+        runs-on: ${{ matrix.os }}
         strategy:
-            max-parallel: 6
+            fail-fast: false
             matrix:
+                os: [ubuntu-latest, macos-latest, windows-latest]
                 emacs-version:
                     - 26.3
                     - 27.2
+                    - 28.1
                     # - snapshot
 
         steps:
             - uses: actions/checkout@v2
 
-            - uses: purcell/setup-emacs@master
+            - uses: jcs090218/setup-emacs@master
               with:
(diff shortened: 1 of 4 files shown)


brotzeit:
Thanks for the great work! cask has been a pain in the ass for quite some time. Did you already announce the package it on reddit ?

jcs090218:
Thanks! :)

No, I haven't announced the package anywhere. This isn't a standard package so it will not be published to MELPA like `Cask` and `Eldev`. :(

But feel free to announce the tool anywhere if you like it! 😄

brotzeit:
I just think there are many package maintainers who would be thankful to have a replacement for cask. But people will start using it for their packages when they see it's used for popular packages like lsp-mode.

jcs090218:
That's basically the goal for Eask; maybe an alternative or replacement to Cask. 😁

PR #458, opened 2022-08-29
Avoid compilation warnings about potentially-undefined functions.

phst:
The byte compiler isn’t smart enough to figure out that these functions are
indeed defined in this code branch.


brotzeit:
Sorry, forgot to merge.

PR #459, opened 2022-09-16
Skip test if Cargo isn’t installed.

phst:
This test requires Cargo and fails if it’s not installed.

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -3482,6 +3482,7 @@ impl Two<'a> {
    "let x = dbg!(\"foo, bar\")"))
 
 (ert-deftest rust-test-project-located ()
+  (skip-unless (executable-find rust-cargo-bin))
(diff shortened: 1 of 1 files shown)


brotzeit:
So the test doesn't run. Any idea why ?

phst:
rust-buffer-project invokes Cargo, cf. https://github.com/rust-lang/rust-mode/blob/2d6bc53f526623e585aae5855a81a13b709df97d/rust-cargo.el#L45. If that fails, it signals an error.

brotzeit:
Sorry, wasn't clear enough ;)

I mean the ci doesn't seem to run the test, otherwise it would fail because we don't install cargo.

phst:
The GitHub runners have Cargo installed by default (https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#rust-tools)

brotzeit:
Interesting, thanks!

PR #466, opened 2022-10-31
* rust-mode.el (rust-syntax-propertize): Don't cache rust macro scopes.

### LINKED ISSUE #465: `rust-syntax-propertize` assumes text already has the properties it is responsible for placing
The `rust-syntax-propertize` function is responsible for setting `syntax-table` text properties on a given region of text. However, it fails to do its job properly, because it uses `syntax-ppss`, which assumes that `syntax-table` properties have already been set.

The `syntax-ppss` function distributed with Emacs calls `syntax-propertize-function` to place `syntax-table` properties as needed, so that when `rust-match-angle-brackets` is enabled, `parse-partial-sexp` can distinguish between `<` and `>` characters surrounding generic parameters and those that are comparison operators or `->` or `=>` tokens. When they enclose generic parameters, `syntax-ppss` treats `<` and `>` tokens as opening and closing "parentheses", in the sense used by Emacs syntax table code.

However, in `rust-mode` buffers, `syntax-propertize-function` is `rust-syntax-propertize`, which ends up calling `syntax-ppss` itself. Unbounded recursion is prevented by `syntax-propertize` temporarily binding `syntax-propertize--done` to `most-positive-fixnum`, but this also means that the propertization doesn't get done.

I don't understand exactly how this happens yet, but I have verified that `parse-partial-sexp` is being called from `rust-syntax-propertize` on regions of text whose `->` tokens have not been marked with `syntax-table` text properties.

jimblandy:
Fixes #465.

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -1411,34 +1411,32 @@ whichever comes first."
 
 (defun rust-syntax-propertize (start end)
   "A `syntax-propertize-function' to apply properties from START to END."
-  ;; Cache all macro scopes as an optimization. See issue #208
-  (let ((rust-macro-scopes (rust-macro-scope start end)))
-    (goto-char start)
-    (let ((str-start (rust-in-str-or-cmnt)))
-      (when str-start
-        (rust--syntax-propertize-raw-string str-start end)))
-    (funcall
-     (syntax-propertize-rules
-      ;; Character literals.
-      (rust--char-literal-rx (1 "\"") (2 "\""))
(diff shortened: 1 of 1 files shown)


jimblandy:
This is obviously not a great fix. But maybe it will be useful to others until something better comes along.

jimblandy:
This is, as you might expect, as slow as paste.

jimblandy:
Withdrawn in favor of #467.

PR #470, opened 2022-11-19
Support envrc for rustfmt

amesgen:
I don't have rustfmt installed globally, but rather get it per-project via Nix + direnv.

A more principled solution would be [inheritenv](https://github.com/purcell/inheritenv), but I think this is simple enough.

Similar to #447.

Less noisy diff due to difftastic:


brotzeit:
Sorry for the delay.

Issue #478, opened 2022-12-20
Window configuration changes on format


ntBre:
I just updated rust-mode and noticed that calling `rust-format-buffer` is now changing my window configuration, namely deleting one of my windows. This appears to be a result of #476 because reverting that change fixes my issue. 

Personally, I think that issue is better fixed by setting `rust-format-show-buffer` to nil, which prevents the extra window from being created in the first place.

Otherwise, if I'm following the code correctly, the `kill-buffer-and-window` call should only be in the `(= ret 3)` case right? I think the current code is assuming that the 0 return value is coming after a non-zero return, which is what created the window in the first place. So if you really want to keep the `kill-buffer-and-window` call, it needs to be gated behind a check for an existing error window, like in the Go code linked in the original issue (#475).

Replacing the kill-buffer call to one adapted from the Go version also seems to work:

```emacs-lisp
(let ((win (get-buffer-window rust-rustfmt-buffername)))
  (if win
      (quit-window t win)
    (kill-buffer rust-rustfmt-buffername)))
```

imichael2e2:
@ntBre Hi!

I think we can discuss more details about your explanation above: >
> Otherwise, if I'm following the code correctly, the
kill-buffer-and-window call should only be in the (= ret 3) case right?
I think the current code is assuming that the 0 return value is coming
after a non-zero return, which is what created the window in the first
place

The `kill-buffer-and-window` call aims to delete the window that is
created when "rustfmt" program detects the error, that is, when the return value
of "rustfmt" subprocess is any value except 0, including the situation
of `(= ret 3)`.

Adding this line in conditional branch of `(zerop ret)` is because when
you got errors, most of time you are about to fix it, after that fix,
the code in `rust–format-call` will run into `(zerop ret)` conditional
branch, the previously created "\*rustfmt\*" buffer will still exist
then. That's why we need to put the killing window code under the
condition of `(zerop ret)` instead of `(= ret 3)`.

ntBre:
Thanks for the additional explanation! I guess I was off base on that speculation, but I did manage to submit a PR that fixed the issue I was having, so I'll close this too.

imichael2e2:
And about the issue you've mentioned: 

> calling rust-format-buffer is now changing my window configuration,
namely deleting one of my windows. 

In fact, I've tested the solution adopted by "go-mode", that is, do some
checks before deleting the window, it turns out it is somewhat
unnecessary.

This is why: the newly popped up window is dedicated for
[...]

Issue #475, opened 2022-12-18
A Redundant Window Is Created, After 'rustfmt'ed


imichael2e2:
For example, here is an error-syntax source file: 
```rust
xxxfn main() {}
```

Open it with either `emacs` or `emacs-nox`, run `rust-format-buffer` for it, an additional buffer `*rustfmt*` shows the stderr:
```
==================================
xxxfn main() {}
============test.rs===============
error: ...
1 | xxxfn main() {}
  |       ^^^^ expected one of `!` or `::`
============*rustfmt*=============
```

After fixed it,  run `rust-format-buffer` would result in two Window, both of which are showing the same file, like this:
```
==================================
fn main() {}
============test.rs===============
fn main() {}
============test.rs===============
```

The second Window might be **redundant**, hence the below one might be more appropriate: 

```
==================================
fn main() {}



============test.rs===============
```

For any lines of code, if there are some errors coming from `rustfmt`, it is expected that we are receiving that information in a seperate auxiliary Buffer,  `rust-mode` perfectly implements such feature. In addition, it is also expected that, once the errors are fixed and the code being formatted, the Buffer used for printing the previous errors should be closed and so does the corresponding Window, so that that the additional temporary Window will not occupy screen space any more. 

But due to the experiment shown above, it seems like `rustfmt` hasn't implemented such feature. For other language formatting packages, this is implemented, for example for `go-mode` package: https://github.com/dominikh/go-mode.el/blob/master/go-mode.el#L1989

brotzeit:
Thanks.

PR #479, opened 2022-12-20
check for rustfmt window before deleting it, fixes the fix to #475

ntBre:
PR version of the fix I proposed in #478


brotzeit:
Sorry for the delay.

PR #482, opened 2022-12-27
provide alternative rust-mode that derives from rust-ts-mode

brotzeit:


thblt:
I was just looking into a way to complement the new `rust-ts-mode` with `rust-mode` features. Wouldn't it be nicer to separate i those features as a minor mode that could be hooked to `rust-ts-mode`?

brotzeit:
If rust-mode was a new package that's supposed to add features to rust-ts-mode it would make sense to use a minor mode. I think there is no real downside to using a derived mode and this way we don't have to apply any additional changes. We can still change this once treesitter is the standard and the old code is deprecated. For now it seems to be the more convenient solution.

But maybe I'm wrong. Other opinions ? @jimblandy @psibi

thblt:
Ha, I see your point, and backward compatibility is indeed a nice thing :)

Maybe it would be nice if the control variable defaulted to t whenever `rust-ts-mode` is available so the transition will be perfectly transparent:

``` elisp
(defcustom rust-mode-treesitter-derive
  (and (fboundp 'treesit-language-available-p)
       (treesit-language-available-p 'rust))
…                              
```

brotzeit:
I thought about it but the ts mode applies different syntax highlighting. Users will probably complain that we just change the default. Many rust-mode users only want very basic features. I know those people probably won't use emacs master, but I would wait.

jimblandy:
I think we should always feel free to make substantial improvements even when they change behavior. If treesitter is available, reliable, and really provides the improvements it claims to, we should just use it.

creese:
Can this be merged?

uncomfyhalomacro:
hello. will this be merged or are there any more to be done first? thank you!

psibi:
I have done a minor change to this PR to make it work for my setup (Emacs 29.1). I can confirm that it works with my basic testing. I'm planning to test drive this PR for around a week to see if I face any issues and merge it if I see no issues.

If others can test drive this PR meanwhile, that will be helpful.

PR #496, opened 2023-04-26
Don’t duplicate test name.

phst:
Emacs 29 will signal an error in this case in batch mode, and in any case the first test in the duplicate series will never run.


psibi:
Thank you!

PR #500, opened 2023-08-04
Eliminate "Containing expression ends prematurely" error

micl2e2:
While #498 handles dbg! insertion for "empty line" cases, the original `rust-insert-dbg` remains unchanged. Though it can handle most cases where there is a valid s-expression ahead of current position, it cannot handle if there isn't, such as:
```
if let Ok(val) = a {}
____________________^ cursor is here 
```
, where it results in "Containing expression ends prematurely" error, and an incorrect insertion:
```
if let Ok(val) = a {(}
```

This patch attempts to fix this issue. And the result would  be:
```
if let Ok(val) = a {dbg!()}
```

### DIFF
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -3449,13 +3449,31 @@ impl Two<'a> {
      "Foo" font-lock-type-face
      "in" font-lock-keyword-face)))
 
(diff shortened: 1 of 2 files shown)


psibi:
Thank you!

PR #501, opened 2023-08-05
test: Emacs 29.1

jcs090218:
Test `29.1` and `snapshot` but allow failure for the `snapshot`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,32 +1,52 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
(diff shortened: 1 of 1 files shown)


psibi:
I would prefer not to have builds for snapshot, atleast for Windows.

jcs090218:
I prefer to build snapshots so people know whether it works on the latest branch or not. That's why we have the `experimental` flag set up. I am assured the Windows snapshot will be fixed since the other two snapshots (ubuntu and macos) have no error. WDYT? 🤔

psibi:
My reasoning is because I have to spend extra time figuring out from github UI to see which job has failed. Okay, let me give this a try. Thanks for the PR!

Issue #507, opened 2023-11-08
compilation-error-regexp not matching panics


ntBre:
I'm not totally sure this is an issue or caused by this package, per se, but I've been having issues with my compilation links to errors. As you can see in the screenshot below, the warnings are working fine, but the whole panic line is turned into a link, and, more importantly, clicking the link brings up `Find this error in (default thread 'atom_typer::tests::atomtyper' panicked at src/lib.rs` rather than taking me to the file. I thought this might have to do with the "cleaner panic messages" from [Rust 1.73](https://blog.rust-lang.org/2023/10/05/Rust-1.73.0.html), but I still had a stable 1.72.1 version installed, and it had the same problem. I don't remember exactly when this stopped working, but I've just been dealing with it for a while and finally decided to ask for help.

I managed to get the behavior I wanted on the error with

```elisp
(setf
 (alist-get 'rustc compilation-error-regexp-alist-alist)
 '("thread '[^']+' panicked at \\(\\(.*\\):\\([0-9]+\\):\\([0-9]+\\)\\)" 2 3 4 nil 1))
```

but that obviously ruins the matching for warnings and other kinds of errors. I would really appreciate any help with this!

I've also been trying to use `rust-ts-mode` a bit, but I don't see anything in there that should be clashing with this option. I'm happy to provide any additional information that might be needed.

![image](https://github.com/rust-lang/rust-mode/assets/36778786/17d635d7-e892-4ee6-b9b7-34d6bc6ce75f)

luckysori:
Great snippet, @ntBre! I was looking for someone with the same problem as me and someone from `#emacs:matrix.org` led me here.

> I would really appreciate any help with this!

I think I made it work by registering a new entry instead of overriding `rustc`:

```elisp
(defvar rustc-panics-compilation-regexps
  (let
      ((re
        "thread '[^']+' panicked at \\(\\(.*\\):\\([0-9]+\\):\\([0-9]+\\)\\)"))
    (cons re '(2 3 4 nil 1))))
(add-to-list
 'compilation-error-regexp-alist-alist
 (cons 'rustc-panics rustc-panics-compilation-regexps))
(add-to-list 'compilation-error-regexp-alist 'rustc-panics)
```

Thank you for sharing your fix!

ntBre:
Wow, great idea! I'll give this a try tomorrow!

ntBre:
[...]

PR #510, opened 2024-02-11
Match panics in `compilation-mode`

luckysori:
Fixes https://github.com/rust-lang/rust-mode/issues/507.

If we encounter a panic when executing Rust code in `compilation-mode`, we now correctly highlight the location where the panic occurred.

For example:

```
thread 'main' panicked at src/main.rs:2:5:
explicit panic
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Here, `src/main.rs:2:5` is highlighted. The developer is then able to execute `compile-goto-error` to jump to the correct spot in the code.

### DIFF
--- a/rust-compile.el
+++ b/rust-compile.el
@@ -34,6 +34,12 @@ See `compilation-error-regexp-alist' for help on their format.")
   "Specifications for matching code references in rustc invocations.
 See `compilation-error-regexp-alist' for help on their format.")
 
+(defvar rustc-panics-compilation-regexps
+   (let ((re (concat "thread '[^']+' panicked at " rustc-compilation-location)))
+     (cons re '(2 3 4 nil 1)))
+   "Specifications for matching panics in rustc invocations.
+See `compilation-error-regexp-alist' for help on their format.")
+
 ;; Match test run failures and panics during compilation as
 ;; compilation warnings
 (defvar cargo-compilation-regexps
@@ -75,6 +81,9 @@ the compilation window until the top of the error is visible."
      (add-to-list 'compilation-error-regexp-alist 'rustc-colon)
      (add-to-list 'compilation-error-regexp-alist-alist
                   (cons 'cargo cargo-compilation-regexps))
+     (add-to-list 'compilation-error-regexp-alist-alist
+                  (cons 'rustc-panics rustc-panics-compilation-regexps))
+     (add-to-list 'compilation-error-regexp-alist 'rustc-panics)
      (add-to-list 'compilation-error-regexp-alist 'cargo)
      (add-hook 'next-error-hook #'rustc-scroll-down-after-next-error)))
 
--- a/rust-mode-tests.el
+++ b/rust-mode-tests.el
@@ -3600,6 +3600,7 @@ let b = 1;"
     (insert "warning found a -> b\n  --> file3.rs:12:34\n\n")
     (insert "note: `ZZZ` could also refer to the constant imported here -> b\n  --> file4.rs:12:34\n\n")
     (insert "    ::: file5.rs:12:34\n\n")
+    (insert "thread 'main' panicked at src/file7.rs:12:34:\n\n")
     ;; should not match
     (insert "werror found a -> b\n  --> no_match.rs:12:34\n\n")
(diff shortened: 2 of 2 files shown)


rust-compile.el
```diff
@@ -34,6 +34,12 @@ See `compilation-error-regexp-alist' for help on their format.")
   "Specifications for matching code references in rustc invocations.
 See `compilation-error-regexp-alist' for help on their format.")
 
+(defvar rustc-panics-compilation-regexps
+   (let ((re "thread '[^']+' panicked at \\(\\(.*\\):\\([0-9]+\\):\\([0-9]+\\)\\)"))
```
zachlefevre, line 38: I think you should use the `rustc-compilation-location` variable already in scope.
Something like this
```
(let ((re (concat "thread '[^']+' panicked at " rustc-compilation-location)))
    (cons re '(2 3 4 nil 1)))
    ```

Comment (luckysori) [reply rust-compile.el]: Done, you were right!

File: rust-mode-tests.el [hunk 2]
```diff
@@ -3618,11 +3619,13 @@ let b = 1;"
                 ("file6.rs" "132" "34" compilation-error "file6.rs:132:34"))
                (("file5.rs" "12" "34" compilation-info "file5.rs:12:34"))
                ((like-previous-one "82" back-to-indentation compilation-info "82")
-                (like-previous-one "132" back-to-indentation compilation-info "132")))
+                (like-previous-one "132" back-to-indentation compilation-info "132"))
+               (("src/file7.rs" "12" "34" nil "src/file7.rs:12:34")))
```
luckysori, line 3623: I've added a test now, but I'm not sure why this is `nil` and not `compilation-error`. Maybe someone more comfortable with Elisp can assist me.

zachlefevre: I am not very familiar with elisp either, so I could be wrong.
Here's what I think is happening here:
compilation-error is set
```
(defvar compilation-error "error"
  "Stem of message to print when no matches are found.")
```
And so an expression like `("file6.rs" "132" "34" compilation-error "file6.rs:132:34")` is equivalent to `("file6.rs" "132" "34" error "file6.rs:132:34")'`

That translates to `type` expression from the `compilation-error-regexp-alist` help doc. I'm getting that from this line
```
Each elt has the form (REGEXP FILE [LINE COLUMN TYPE HYPERLINK
HIGHLIGHT...]).  If REGEXP matches, the FILE’th subexpression
gives the file name, and the LINE’th subexpression gives the line
number.  The COLUMN’th subexpression gives the column number on
that line.
```
The definition of our regex is
```
defvar rustc-panics-compilation-regexps
   (let ((re (concat "thread '[^']+' panicked at " rustc-compilation-location)))
     (cons re '(2 3 4 nil 1)))
   "Specifications for matching panics in rustc invocations.
See `compilation-error-regexp-alist' for help on their format.")
```
And so we're capturing the:
2nd capture group for the file name
3rd capture group for the line number,
4th capture group for the column number,
nil for the type
1st capture group for the hyperlink.

We don't have an `error` or `warning` in our error message, like the others do, to parse out for the `type`, so our `type` is `nil`.

zachlefevre: so I think `nil` is correct. That, or invent a new type like `panic`. I'm not really sure.

zachlefevre:
@luckysori This looks ready for review to me. Do you mind running `@rustbot review`? from https://rustc-dev-guide.rust-lang.org/contributing.html#waiting-for-reviews ?

luckysori:
> **Error**: The feature `shortcut` is not enabled in this repository. To enable it add its section in the `triagebot.toml` in the root of the repository.
> 
> Please file an issue on GitHub at [triagebot](https://github.com/rust-lang/triagebot) if there's a problem with this bot, or reach out on [#t-infra](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra) on Zulip.

I guess it's not enabled on this repo, @zachlefevre.

zachlefevre:
@psibi Do you mind reviewing this?

psibi:
Thanks @zachlefevre for the review and @luckysori for the PR and the test case! I have confirmed that it works for me, merging this.

PR #511, opened 2024-02-12
Add compilation regex for panics.

zachlefevre:
This matches errors in the compilation buffer like
```
thread 'main' panicked at src/main.rs:16:5:
```
and
```
thread 'foo' panicked at tests/health_check.rs:3:5:
```


zachlefevre:
duplicate of #510

PR #519, opened 2024-02-27
Fix undefined function

### LINKED ISSUE #516: Cannot setup LSP formatting and treesitter together
Hello,

Here is my current config:

```elisp
(use-package rust-mode
  :hook (rust-mode . lsp-deferred)
  :init
  (setq rust-mode-treesitter-derive t)
  :config
  (setq rust-format-on-save t
        rust-after-save-method 'lsp-format-buffer))
```

Then in rust-mode (when editing a Rust file), the format on save is not happening.

Strangely, it says:
```
run-hooks: Symbol’s function definition is void: rust-after-save-method
```
I keep having the same message if I try too change what is in the `:init` and `:config` statements.

ccqpein:
Try to fix #516

### DIFF
--- a/rust-common.el
+++ b/rust-common.el
@@ -4,6 +4,8 @@
 ;; rust-common code for both prog-mode and tree-sitter one
 
 ;;; Code:
+(require 'rust-rustfmt)
+
 (defcustom rust-before-save-hook 'rust-before-save-method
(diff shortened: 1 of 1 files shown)


psibi:
Thank you!

PR #521, opened 2024-02-28
Fix tree-sitter integration for older emacs

psibi:


alanz:
Thanks for the quick fix. This sorted out the unknown hooks, but I now get

```
Warning (treesit): Cannot activate tree-sitter, because Emacs is not compiled with tree-sitter library
```
I am running emacs 29.2 but without tree-sitter integration.

psibi:
@alanz Can you see if this PR fixes it for you: https://github.com/rust-lang/rust-mode/pull/522 ?

PR #522, opened 2024-02-28
Enable tree sitter code only when it's enabled

psibi:


alanz:
This does not work for me. When I apply it locally, flycheck reports that `rust-mode-treesitter-derive)` is not defined

alanz:
Perhaps that should be an or?

alanz:
Adding `(require 'rust-mode)` gets rid of the warning, but the problem remains.

Turning the `and` into an `or` just makes rust-mode not load at all.

I am reluctant to turn on native tree-sitter mode, as I develop some tree-sitter grammars and have the non-native tree-sitter support enabled.

psibi:
@alanz When do you actually see that warning that was reported by you ? Is it during installation or during some other step (Your latest comment indicates it's during flycheck) ? I can try reproducing it locally for my setup.

I don't think doing `or` is the right thing here.

alanz:
I agree or is not the right step, it broke completely.

I see the error on starting up emacs.  I guess I need to dig more deeply into the actual origin. And/or go full native tree-sitter.
I will not be able to do any investigation today though

alanz:
And it seems the warning is benign, I have the functionality I need regardless.

psibi:
>  see the error on starting up emacs.

Looking at my Messages buffer - I can see a similar type of warning:

```
 ■  Warning (treesit): Cannot activate tree-sitter, because language grammar for rust is unavailable (not-found): (libtree-sitter-rust libtree-sitter-rust.0 libtree-sitter-rust.0.0 libtree-sitter-rust.so libtree-sitter-rust.so.0 libtree-sitter-rust.so.0.0) No such file or directory
```

I receiving the warning even though I'm using a tree sitter based setup. Probably it's able to find the grammars subsequently, not sure.

>  And it seems the warning is benign, I have the functionality I need regardless.

Thanks for the confirmation, I assumed something was broken. Given the situation, I will close this PR. I would be happy to accept any PR that fixes the warning. :-)

alanz:
FYI, I switched to emacsformacosx, which has tree-sitter support build in. Without explicitly asking for the tree-sitter mode, I am getting

```
⛔ Warning (treesit): Cannot activate tree-sitter, because language grammar for rust is unavailable (not-found): (libtree-sitter-rust.so libtree-sitter-rust.so.0 libtree-sitter-rust.so.0.0 libtree-sitter-rust.dylib libtree-sitter-rust.dylib.0 libtree-sitter-rust.dylib.0.0) No such file or directory
```

Which I guess means I need to work out how to install the language. Is it documented somewhere? A quick scan of the README does not show anything

alanz:
And it looks like the steps are

init.el
```elisp
  (setq treesit-language-source-alist
      '((rust "https://github.com/tree-sitter/tree-sitter-rust")
        ))
```

In emacs
`M-x` treesit-install-language-grammar`
Pick `rust`

alanz:
My final piece of barrage.

Putting the following into my init.le

```elisp
(use-package rust-ts-mode
  :custom
  (rust-mode-treesitter-derive t)
...
)
...
(use-package lsp-mode
  :hook (rust-mode rust-ts-mode)
...
)
```
And basically using `rust-ts-mode` instead of `rust-mode` in all config restored
- semantic highlighting
- compile buffer regexp matching
- basic rust-mode operation, such as knowledge of what the normal indent and comment markers are

PR #530, opened 2024-03-08
Fix rust-mode lazy loading

### LINKED ISSUE #528: File mode specification error: (void-function rust-mode)
The following error is returned when opening a `.rs` file:

```
File mode specification error: (void-function rust-mode)
```

It seems to be caused by:

https://github.com/rust-lang/rust-mode/blob/d8a09f218e24407acbc9f36c641be4f913f1a63c/rust-mode.el#L78-L79

Because after the last changes to support treesitter, no `rust-mode` is autoloaded.

https://github.com/rust-lang/rust-mode/blob/d8a09f218e24407acbc9f36c641be4f913f1a63c/rust-mode.el#L74-L76

https://github.com/rust-lang/rust-mode/blob/d8a09f218e24407acbc9f36c641be4f913f1a63c/rust-prog-mode.el#L1448

https://github.com/rust-lang/rust-mode/blob/d8a09f218e24407acbc9f36c641be4f913f1a63c/rust-mode-treesitter.el#L15

My current workaround it is to `require 'rust-mode` in my `init.el`. However, this was not necessary before and it woud be great to continue supporting lazy-loading.

jroimartin:
This PR fixes the following error that happends when opening a .rs
file:

	File mode specification error: (void-function rust-mode)

It conditionally autoloads the proper rust-mode version depending on
the user environment.

Fixes #528

### DIFF
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -76,6 +76,9 @@ instead of `prog-mode'. This option requires emacs29+."
   (require 'rust-prog-mode))
 
 ;;;###autoload
+(autoload 'rust-mode "rust-mode" "Major mode for Rust code.")
+
+;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.rs\\'" . rust-mode))
 
 (provide 'rust-mode)


psibi:
Thanks, @jroimartin can you see why the tests are failing ?

jroimartin:
Sure, I'm working on it.

jroimartin:
@psibi PTAL. I think this is the smallest required change. On the flip side, it is a bit tricky. But having two major modes with the same name and loading them conditionally seems problematic. So, I opted for allowing the stub function (called `rust-mode`) to be redefined when loading the actual mode.

psibi:
Thanks, LGTM. @condy0919 Can you see if it works in your setup too ? I will test drive this PR too.

condy0919:
> Thanks, LGTM. @condy0919 Can you see if it works in your setup too ? I will test drive this PR too.

Sorry, I'm in a touring🥺 can't help that

jroimartin:
No rush. There is a valid workaround which is just adding `(require 'rust-mode)` to `init.el`. It is just that it is nice to support lazy loading :slightly_smiling_face: This change also make `rust-mode` just work for people using plain `package.el`.

psibi:
No worries, I'm also going to be on travel for the next two days. I will test this out more properly after that.

rust-mode.el
```diff
@@ -71,9 +71,15 @@ instead of `prog-mode'. This option requires emacs29+."
     map)
   "Keymap for Rust major mode.")
 
-(if (and (version<= "29.1" emacs-version) rust-mode-treesitter-derive)
-    (require 'rust-mode-treesitter)
-  (require 'rust-prog-mode))
+;;;###autoload
+(defun rust-mode ()
```
psibi, line 75: Doesn't this function conflict with the `rust-mode` function that is exposed by the rust-prog-mode / rust-treesitter-mode ?

psibi: Infact, give that you are calling `rust-mode` in the body of this function - it makes me more nervous. Do you think we can rename it to a different function instead, probably `rust-mode-load` or something like that ?

jroimartin: @psibi yeah, it is a bit hacky...

The thing is that we want to autoload `rust-mode.el` when the user calls `(rust-mode)` and then select the actual major mode. However, both versions of the major mode (`rust-prog-mode` and `rust-treesitter-mode`) are called `rust-mode` and which one to choose depends on the environment (the emacs version and the value of `rust-mode-treesitter-derive`). So, I don't know how to autoload one or the other conditionally.

Also, we cannot change the name of the major modes (and autoload both) because the hooks are named after them and other extensions expect these modes to be called `rust-mode`.

So, this PR autoloads a stub called `rust-mode` that is redefined by the specific call to `define-derived-mode`. Thus, the following calls to `rust-mode` execute the `rust-mode` function corresponding to the actual major mode.

condy0919: > probably `rust-mode-load` or something like that ?

Or `rust-mode-maybe`, it's mentioned in the corresponding issue. But using another name may break users' config as many people manually do

```elisp
(add-to-list 'auto-mode-alist '("\\.rs\\'" . rust-mode))
```

jroimartin:
@psibi @condy0919 what do you think of db7d086? This option does not require messing up with the `rust-mode` function definition.

condy0919:
> @psibi @condy0919 what do you think of db7d086? This option does not require messing up with the `rust-mode` function definition.

This solution comes to mind at first, and the trick is used in the evil-exchange package.

psibi (approved):

psibi:
I did some testing and this seems to work fine. @jroimartin Thanks for the PR. @condy0919 Thanks for testing and review.

PR #532, opened 2024-03-12
Allow autoloaded rust-mode function to be called interactively

jroimartin:
This PR updates the autoload call so the rust-mode function can be
called interactively (using `M-x`).


psibi (approved):

psibi:
Thank you!

Issue #527, opened 2024-03-05
Remove autoload might break the `straight-use-package-mode`


ccqpein:
I found the latest `rust-mode` cannot load successfully by my `use-package` (turned on the straight `use-package` integrated).

After some investigating, I found the PR #526 remove the `;;;###autoload` of `rust-mode`. It causes `straight-use-package` cannot find the `rust-mode` function. (I guess it doesn't load the `rust-mode.el` when it initing?)

After more playing around, there are several behaviors were revealed. 

1. if I `(load "~/.emacs.d/straight/repos/rust-mode/rust-mode.el")` in `*scratch*` buffer. I can call `M-x rust-mode`. But `M-x straight-use-package` cannot find the `rust-mode`
2. if I `(autoload 'rust-mode "rust-mode" nil t)` in `*scratch*` buffer. It does the same thing as `1.`.
3. If I load `rust-mode.el` in `use-package` `:init` step. It works fine. Furthermore, if I `(setq rust-mode-treesitter-derive t)` before in `:init` before loading, it will in treesitter mode as master branch code is designed to do. 

I haven't raised up the solution of this issue. I guess it is the `autoload` is necessary for `straight.el`. I am going to report this to them.

ccqpein:
Also raised the question to `straight.el` repo. [link](https://github.com/radian-software/straight.el/issues/1152)

jroimartin:
Related: #528

condy0919:
```elisp
;; rust-mode.el ...

;;;###autoload
(autoload 'rust-mode "rust-mode.el" nil t)
```

It will generate an `(autoload 'rust-mode "rust-mode.el" nil t)` in rust-mode-autoloads.el, it works well. But it looks like weird as of an autoload is autoloaded.

psibi:
@ccqpein Can you check if this PR by @jroimartin fixes it for you: https://github.com/rust-lang/rust-mode/pull/530 ?

ccqpein:
@psibi I just try and it fixes the `rust-mode`. Now my rust file can load the `rust-mode` (from `rust-prog-mode.el`). However, the `straight-use-package` still cannot find the `rust-mode` by `M-x straight-use-package`. 

I might find sometime to study how the `straight.el` handle `straight-use-package`

ccqpein:
Figure out it works. Thanks for @jroimartin 's PR

Issue #520, opened 2024-02-27
Compiling/Installing error: No such file or directory, treesit


KDr2:
- OS: Linux (Ubuntu 22.04)
- Emacs Version: 28.2

I got an error while installing `rust-mode`:
```
$HOME/.emacs.d/elpa/rust-mode-20240225.526/rust-mode-treesitter.el at Tue Feb 27 21:04:29 2024
rust-mode-treesitter.el:9:1:Error: Cannot open load file: No such file or directory, treesit
```
it's caused by this line: https://github.com/rust-lang/rust-mode/blob/master/rust-mode-treesitter.el#L9.

The problem doesn't exist in Emacs 29.1.

Does anyone know how to fix this? Thank you.

psibi:
How does your rust-mode configuration look like ?

psibi:
Although the error is surprising since that code should not get evaluated for older emacs: https://github.com/rust-lang/rust-mode/blob/35e68c30fc882dd03c562346fa81eed6c7e09a2f/rust-mode.el#L75

KDr2:
I started emacs with `emacs -q` and ran the code below manually:

```
(require 'package)
(add-to-list 'package-archives
             '("melpa" . "https://melpa.org/packages/") t)
(package-initialize)
(package-refresh-contents)
(package-install 'rust-mode)
```

Then the error occurred.

KDr2:
The complete log is like this:

```
Leaving directory ‘$HOME/.emacs.d/elpa/rust-mode-20240225.526’
^L
Compiling file $HOME/.emacs.d/elpa/rust-mode-20240225.526/rust-cargo.el at Tue Feb 27 21:04:29 2024
Entering directory ‘$HOME/.emacs.d/elpa/rust-mode-20240225.526/’
^L
Compiling file $HOME/.emacs.d/elpa/rust-mode-20240225.526/rust-common.el at Tue Feb 27 21:04:29 2024
^L
Compiling file $HOME/.emacs.d/elpa/rust-mode-20240225.526/rust-compile.el at Tue Feb 27 21:04:29 2024
^L
Compiling file $HOME/.emacs.d/elpa/rust-mode-20240225.526/rust-mode-treesitter.el at Tue Feb 27 21:04:29 2024
rust-mode-treesitter.el:9:1:Error: Cannot open load file: No such file or directory, treesit
^L
[...]

PR #534, opened 2024-03-16
Fix ancient defect and dependency complications related to having two rust-mode implementations

tarsius:
```
lib/rust-mode/rust-mode-treesitter.el:21:33: Warning: reference to free variable ‘rust-before-save-hook’
lib/rust-mode/rust-mode-treesitter.el:22:32: Warning: reference to free variable ‘rust-after-save-hook’
```

and

```
lib/rust-mode/rust-rustfmt.el:60:12: Warning: ‘unwind-protect’ without unwind forms
```

### DIFF
--- a/rust-mode-treesitter.el
+++ b/rust-mode-treesitter.el
@@ -5,12 +5,13 @@
 
 ;;; Code:
 
+(require 'rust-mode)
+
+;; Do not compile or load on Emacs releases that don't support
+;; this.  See https://github.com/rust-lang/rust-mode/issues/520.
 (when (version<= "29.1" emacs-version)
-  ;; We have the when macro because of
-  ;; https://github.com/rust-lang/rust-mode/issues/520
   (require 'treesit)
   (require 'rust-ts-mode)
-  (require 'rust-common)
 
   (define-derived-mode rust-mode rust-ts-mode "Rust"
     "Major mode for Rust code.
--- a/rust-mode.el
+++ b/rust-mode.el
@@ -71,10 +71,6 @@ instead of `prog-mode'. This option requires emacs29+."
     map)
   "Keymap for Rust major mode.")
 
-(if (and (version<= "29.1" emacs-version) rust-mode-treesitter-derive)
-    (require 'rust-mode-treesitter)
-  (require 'rust-prog-mode))
-
 ;;;###autoload
 (autoload 'rust-mode "rust-mode" "Major mode for Rust code." t)
 
@@ -82,6 +78,12 @@ instead of `prog-mode'. This option requires emacs29+."
 (add-to-list 'auto-mode-alist '("\\.rs\\'" . rust-mode))
 
 (provide 'rust-mode)
+
+(if (and rust-mode-treesitter-derive
+         (version<= "29.1" emacs-version))
+    (require 'rust-mode-treesitter)
+  (require 'rust-prog-mode))
+
 (require 'rust-utils)
 
 ;;; rust-mode.el ends here
--- a/rust-prog-mode.el
+++ b/rust-prog-mode.el
@@ -4,7 +4,8 @@
 ;; rust-mode code deriving from prog-mode instead of rust-ts-mode
 
 ;;; Code:
-(require 'rust-common)
+
+(require 'rust-mode)
 
 (defvar electric-pair-inhibit-predicate)
 (defvar electric-pair-skip-self)
--- a/rust-rustfmt.el
+++ b/rust-rustfmt.el
@@ -87,8 +87,8 @@
                 (insert-file-contents tmpf)
                 (rust--format-fix-rustfmt-buffer (buffer-name buf))
                 (error "Rustfmt failed, see %s buffer for details"
-                       rust-rustfmt-buffername))))
-          (delete-file tmpf))))))
+                       rust-rustfmt-buffername)))
+            (delete-file tmpf)))))))
 
 ;; Since we run rustfmt through stdin we get <stdin> markers in the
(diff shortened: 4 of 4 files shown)


rust-mode-treesitter.el
```diff
@@ -5,6 +5,9 @@
 
 ;;; Code:
 
+(defvar rust-before-save-hook)
```
psibi, line 8: Shouldn't this come from `rust-common.el` ?

tarsius: Yes, but it seems you forgot to explicitly require that. It appears that, for some reason, which escapes me now, I assumed there was a reason why this library doesn't require any other libraries.

psibi: But `rust-mode.el` does require it: https://github.com/rust-lang/rust-mode/blob/bd7d8ed2ba324dce7e531134d405eb4a8c14d876/rust-mode.el#L17

tarsius: But `rust-mode-treesitter.el` does not require `rust-mode` or any other library.

psibi: That's because `rust-mode.el` requires `rust-mode-treesitter.el`: https://github.com/rust-lang/rust-mode/blob/bd7d8ed2ba324dce7e531134d405eb4a8c14d876/rust-mode.el#L74

tarsius: And *that* is why we need to declare this variable. When we compile `rust-mode-treesitter.el`, the compiler doesn't know that at runtime, `rust-mode-treesitter` is loaded because it is required in `rust-mode`, and that it is assumed the user (or some third-party code) does *not* load/require `rust-mode-treesitter` directly. So we have to tell the compiler about it, so it can stop being alarmist about it, and so that we don't get used to it outputting warnings and thus end up ignoring them.

However, I don't think that is right way to do it. This mode *does* depend other parts of this package, so it should depend on them using `require`. This is complicated by the fact that you also want to load this file, whenever `rust-mode` is loaded (modulo some conditions, that are not relevant to this discussion).

You *can* have it both ways, by moving `(require 'rust-mode-treesitter)` below `(provide 'rust-mode)` (as is already being done for `(require 'rust-utils)`).

tarsius: `rust-prog-mode` should also require `rust-mode`.

psibi: > You can have it both ways, by moving (require 'rust-mode-treesitter) below (provide 'rust-mode) (as is already being done for (require 'rust-utils)).

Can you update your PR like that ? This whole require thing has caused many issues that it makes me nervous in general. Eg: https://github.com/rust-lang/rust-mode/issues/520, https://github.com/rust-lang/rust-mode/issues/527

tarsius: I've pushed an update.

psibi: Thanks, I did a quick testing and this it looks good. I will use it throughout the day and see if any issue pops up.

psibi:
@jroimartin / @condy0919 Can you also test drive this PR since you have had related issues previously ?

condy0919:
> @jroimartin / @condy0919 Can you also test drive this PR since you have had related issues previously ?

My pleasure.

jroimartin:
> @jroimartin / @condy0919 Can you also test drive this PR since you have had related issues previously ?

@psibi it seems to work fine on my side.

rust-mode.el
```diff
@@ -71,17 +71,19 @@ instead of `prog-mode'. This option requires emacs29+."
...
 ;;;###autoload
 (autoload 'rust-mode "rust-mode" "Major mode for Rust code." t)
 
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.rs\\'" . rust-mode))
 
 (provide 'rust-mode)
+
+(if (and rust-mode-treesitter-derive
+         (version<= "29.1" emacs-version))
```
sebastiencs, line 83: Emacs > 29.1 can be compiled without tree sitter feature.
We should maybe replace this with `(functionp 'treesit-parser-p)`, or something similar ?

psibi: That's why this is opt in feature, by default `rust-mode-treesiter-derive` is `nil`. 

> We should maybe replace this with (functionp 'treesit-parser-p), or something similar ?

Even then it should be opt-in IMO (atleast for a couple of years) and that already introduces a new defcustom variable. Also, in my experience I have found `treesit-parser-p` not reliable where during startup it returned nil.

sebastiencs: I mean we should replace `(version<= "29.1" emacs-version)` with `(functionp 'treesit-parser-p)`.

>  Also, in my experience I have found treesit-parser-p not reliable where during startup it returned nil.

`(functionp 'treesit-parser-p)` doesn't call `treesit-parser-p`, it will just return nil when emacs has not been compiled with treesit (`--without-tree-sitter` flag at compile time)

psibi: > I mean we should replace (version<= "29.1" emacs-version) with (functionp 'treesit-parser-p).

Ah, I see what you mean. Probably you can open a separate PR for it ?

condy0919: And also the `if` check in rust-mode-treesitter.el

psibi:
I'm merging this since I have seen no objections from others. Thanks @tarsius for the PR! Thanks @condy0919 and @jroimartin for the additional testing.

rust-mode.el
psibi: Yeah, agree. I have opened an issue to track this: https://github.com/rust-lang/rust-mode/issues/537

tarsius:
You're welcome!

PR #538, opened 2024-04-13
fix: move indent level to rust-common

Dev380:
Emacs complains that `rust-indent-offset`
isn't set and a few minor modes (e.g.
formatting, smartparens) stop working.
This moves that variable to the common
file so the treesitter mode has the
variable set.

I'm not sure why more of the rust-prog-mode variables aren't in the rust-common file so I left them there. But, this was the only variable that was causing breakage for me. There is a rust-ts-mode-indent-offset variable too (which is also 4 by default) - should rust-mode override that variable with this one, override the rust-mode variable with the tree sitter one when the tree sitter mode is on, or keep them independent?

### DIFF
--- a/rust-common.el
+++ b/rust-common.el
@@ -16,5 +16,11 @@
   :type 'function
   :group 'rust-mode)
 
+(defcustom rust-indent-offset 4
+  "Indent Rust code by this number of spaces."
+  :type 'integer
+  :group 'rust-mode
+  :safe #'integerp)
+
 (provide 'rust-common)
 ;;; rust-common.el ends here
--- a/rust-prog-mode.el
+++ b/rust-prog-mode.el
@@ -17,12 +17,6 @@
     ("INFINITY" . ?∞) ("->" . ?→) ("=>" . ?⇒))
   "Alist of symbol prettifications used for `prettify-symbols-alist'.")
 
-(defcustom rust-indent-offset 4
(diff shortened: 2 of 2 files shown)


rust-common.el
```diff
@@ -16,5 +16,11 @@
   :type 'function
   :group 'rust-mode)
 
+(defcustom rust-indent-offset 4
```
psibi, line 19: Instead of moving it here, do you think a defvaralias should be instead created. Something similar to this: https://github.com/psibi/rustic/blob/4dac7f34c30d91e3ecb2602fff7e81b6efdc75d7/rustic.el#L52

psibi:
@Dushistov Does this PR fix that for you ?

Dushistov:
> @Dushistov Does this PR fix that for you ?

It fixes, but for reason introduce new bug: lsp-mode for files except `lib.rs` stop detects that it belong to project.

So for  `file.rs` lsp-mode not activated and I have to run M-x lsp-mode for all files except lib.rs.

psibi:
> It fixes, but for reason introduce new bug: lsp-mode for files except lib.rs stop detects that it belong to project.

That strange and looks to be a different bug since lsp-mode detection is not done in rust-mode AFAIK.

If you can address this comment: https://github.com/rust-lang/rust-mode/pull/538#discussion_r1565482482 I think we can merge a new PR.

PR #544, opened 2024-05-13
Support running interactive programs

isvilen:
If universal argument is set before invoking `rust-run`  `(C-u C-c C-c C-r)` the output buffer will be in `Comint` mode with
`compilation-shell-minor-mode` and user input can be processed.

### DIFF
--- a/rust-cargo.el
+++ b/rust-cargo.el
@@ -67,46 +67,54 @@
 
 ;;; Internal
 
-(defun rust--compile (format-string &rest args)
+(defun rust--compile (comint format-string &rest args)
   (when (null rust-buffer-project)
     (rust-update-buffer-project))
   (let ((default-directory
           (or (and rust-buffer-project
                    (file-name-directory rust-buffer-project))
-              default-directory)))
-    (compile (apply #'format format-string args))))
+              default-directory))
+        ;; make sure comint is a boolean value
+        (comint (not (not comint))))
+    (compile (apply #'format format-string args) comint)))
 
 ;;; Commands
 
 (defun rust-check ()
   "Compile using `cargo check`"
   (interactive)
-  (rust--compile "%s check %s" rust-cargo-bin rust-cargo-default-arguments))
+  (rust--compile nil "%s check %s" rust-cargo-bin rust-cargo-default-arguments))
 
 (defun rust-compile ()
   "Compile using `cargo build`"
   (interactive)
-  (rust--compile "%s build %s" rust-cargo-bin rust-cargo-default-arguments))
+  (rust--compile nil "%s build %s" rust-cargo-bin rust-cargo-default-arguments))
 
 (defun rust-compile-release ()
   "Compile using `cargo build --release`"
   (interactive)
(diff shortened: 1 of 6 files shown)


rust-cargo.el
```diff
@@ -67,46 +67,52 @@
...
-(defun rust--compile (format-string &rest args)
+(defun rust--compile (comint format-string &rest args)
   (when (null rust-buffer-project)
     (rust-update-buffer-project))
   (let ((default-directory
           (or (and rust-buffer-project
                    (file-name-directory rust-buffer-project))
               default-directory)))
-    (compile (apply #'format format-string args))))
+    (compile (apply #'format format-string args) (not (not comint)))))
```
psibi, line 77: Isn't `(not (not comint))` is same as `comint` ?

psibi:
@isvilen Looks good to me apart from minor comment. But for this to be merged, this would need two more updates:

* Update the documentation for this in README
* Add a specific test for this feature.

rust-cargo.el
isvilen: Not in ELisp :), this is needed to ensure that `comint` argument is a boolean value, so that it works with
universal argument provided by `C-u`. I've added a comment about that in the source.

isvilen:
I've updated the README file, but I can find any existing tests for `rust-cargo.el`, so I am not sure how to add a test for this feature.

psibi:
> so I am not sure how to add a test for this feature.

We have some tests for it in rustic mode for cargo-run etc: https://github.com/emacs-rustic/rustic/blob/17a79c659b0eb304e35705facb114caf61eae2e9/test/rustic-cargo-test.el#L6

Can you see if you can adapt it from there ? It's okay if it's a separate file.

rust-cargo.el
psibi: Ah, got it. Thanks!

isvilen:
@psibi Added  tests for `rust-compile` and `rust-run` in interactive and non-interactive mode.

psibi:
@isvilen Thanks, it works and I tested it locally too with a few different program. 

Although I was not successful with the `test-project/src/main.rs`: How do you ensure when running using `rust-run` that `args.len() != 1` since I don't see a way to pass a parameter to `cargo run` itself. 

Probably I'm missing something since I mostly use `rustic` and `rust-mode` has different way to pass an option ?

isvilen:
> @isvilen Thanks, it works and I tested it locally too with a few different program.
> 
> Although I was not successful with the `test-project/src/main.rs`: How do you ensure when running using `rust-run` that `args.len() != 1` since I don't see a way to pass a parameter to `cargo run` itself.
> 
> Probably I'm missing something since I mostly use `rustic` and `rust-mode` has different way to pass an option ?

I just set `rust-cargo-default-arguments` before invoking `rust-run` see https://github.com/rust-lang/rust-mode/blob/947ebf993818f7135e12d4b0da6f5b05ba790c14/rust-cargo-tests.el#L47

psibi:
Thank you!

Andsbf:
thanks you both for that @psibi @isvilen 💯

Issue #545, opened 2024-05-13
Add configuration option to locate project without --workspace to only run check on local crate


bradneuman:
I'm working in a project that uses a Cargo Workspace and has some large crates. I want to be able to iterate with commands like `cargo check` on just my crate.

I think one wait to do this would be to add a config to not include `--workspace` in the `cargo locate-project` command.

I might take a stab at it, but wanted to post here in case others are interested or there's a better workaround already

bradneuman:
OK took a crack at it, seems to work for me

psibi:
Thanks, I saw your PR. I think instead of the approach you have taken here, you should make it accept universal argument to pass custom arguments instead.

That's how we do it in rustic where this workflow is already implemented: https://github.com/emacs-rustic/rustic/blob/17a79c659b0eb304e35705facb114caf61eae2e9/rustic-cargo.el#L703

bradneuman:
Ah that makes sense, and then just have it default to `--workspace`?

Although now I might just switch to `rustic` :-D

psibi:
> Ah that makes sense, and then just have it default to --workspace?

This is the default we have there: https://github.com/emacs-rustic/rustic/blob/17a79c659b0eb304e35705facb114caf61eae2e9/rustic-cargo.el#L90

But yeah, `--workspace` also seems a reasonable default.

> Although now I might just switch to rustic :-D

Cool, note that the installation instruction might need some updates etc. Feel free to send PR's. The fork is not officially announced yet. For more background, here is the information: https://psibi.in/posts/rustic.html

psibi:
> whereas rust.el is setting the project root once in the function rust-buffer-project and then using it for each sub-command in rust--compile.

You're right. 

Given the situation, I'm starting to lean towards your current implementation.

> but now I need to start thinking about a migration path

I'm not sure if it's a good idea to cause such a big breaking change.

bradneuman:
That's definitely fine with me :)

I haven't used github in a while and I can't seem to figure out how to kick off the CI job again - though it looks like it should be fixed. Am I missing something obvious? (All the internet instructions just say to click some button with a rewind icon but of course I don't have that button)

psibi:
@bradneuman I have left a comment here: https://github.com/rust-lang/rust-mode/pull/546#discussion_r1600884096

psibi:
Closing this as it has been addressed by https://github.com/rust-lang/rust-mode/pull/546

PR #546, opened 2024-05-13
Adds a customization to default to workspace or local crate.

bradneuman:
Adds rust-locate-project-in-workspace custom variable, which controls whether or not to locate the workspace project using `--workspace` (the default) or not.

In cases where there is only one create, this should make no difference.

The default setting should match the existing behavior.

Github issue #545

### DIFF
--- a/rust-cargo.el
+++ b/rust-cargo.el
@@ -19,11 +19,11 @@
   :type 'boolean
   :group 'rust-mode)
 
-(defcustom rust-locate-project-in-workspace t
-  "Whether to use `--workspace` with `cargo locate-project`. If t,
- rust-mode will run commands for the entire workspace. If nil,
- rust will search for the Cargo.toml in the local crated"
-  :type 'boolean
+(defcustom rust-cargo-locate-default-arguments '("--workspace")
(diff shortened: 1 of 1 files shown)


rust-cargo.el
```diff
@@ -19,6 +19,13 @@
   :type 'boolean
   :group 'rust-mode)
 
+(defcustom rust-locate-project-in-workspace t
+  "Whether to use `--workspace` with `cargo locate-project`. If t,
+ rust-mode will run commands for the entire workspace. If nil,
+ rust will search for the Cargo.toml in the local crated"
+  :type 'boolean
+  :group 'rust-mode)
+
```
psibi, line 28: How about we change this to `rust-cargo-locate-default-arguments` ? It's default value could be `(list "--workspace")`. For your specific use case, you can pass it as `nil`.

This will give you the ability to pass any arbitrary arguments to `locate-project`.

bradneuman: Done

psibi (approved):

psibi:
Thank you!

PR #554, opened 2024-08-31
Update rustfmt's default edition: 2018 -> 2021

PuercoPop:
The latest edition seems like a good default.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Update rustfmt's defaults to use 2021 edition ([#554](https://github.com/rust-lang/rust-mode/issues/509)).
+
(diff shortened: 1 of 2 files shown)


psibi (commented):
Can you also update the changelog with an unreleased section?

PuercoPop:
> Can you also update the changelog with an unreleased section?

Done

psibi (approved):
Thanks!

apiraino:
For the record, this change will be neutralized in https://github.com/rust-lang/rust-mode/pull/571

I reported in https://github.com/rust-lang/rust-mode/issues/570 that rust-rustfmt-switches was shadowing the rustfmt.toml, ignoring any user configuration.

PR #563, opened 2024-11-03
Indent example-code blocks in comments

bemoody:
Allow pressing Tab to indent example code blocks inside comments.

For example, if the buffer contains this:
```
/// Say hello
///
/// ```
/// fn main() {
/// greet("world");
/// }
/// ```
pub fn greet(whom: &str) {
    println!("Hello, {}", whom);
}
```

Move to the word "world" and press Tab, and you should see:
```
/// Say hello
///
/// ```
/// fn main() {
///     greet("world");
/// }
/// ```
pub fn greet(whom: &str) {
    println!("Hello, {}", whom);
}
```

### DIFF
--- a/rust-prog-mode.el
+++ b/rust-prog-mode.el
@@ -769,6 +769,123 @@ buffer."
   (save-excursion (= (progn (goto-char pos1) (line-end-position))
                      (progn (goto-char pos2) (line-end-position)))))
 
+(defun rust-doc-indent-line ()
+  "Indent the current line, and indent code examples in comments.
+
+Indent the current line as `rust-mode-indent-line' does.  If
+point is inside a block comment containing a Markdown code
+example (delimited by triple backquotes), then also indent the
+current line within the code example."
+  (interactive)
+
+  ;; First, reindent the current line.
+  (rust-mode-indent-line)
+
+  ;; If point is inside a comment:
+  (let ((ppss (syntax-ppss)))
+    (when (nth 4 ppss)
+      (rust-with-comment-fill-prefix
+       (lambda ()
+         (let* ((orig-buf (current-buffer))
+                (orig-point (point))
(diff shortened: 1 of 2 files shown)


rust-prog-mode.el
```diff
@@ -1457,7 +1574,7 @@ whichever comes first."
   (setq-local syntax-propertize-function #'rust-syntax-propertize)
 
   ;; Indentation
-  (setq-local indent-line-function 'rust-mode-indent-line)
+  (setq-local indent-line-function 'rust-doc-indent-line)
```
jcs090218, line 1577: Let's keep this variable to `rust-mode-indent-line` and tweak the functionality in that function. 🤔

bemoody:
I'm not sure what you're saying.

I could rename the function rust-mode-indent-line to rust-simple-indent-line and rename rust-doc-indent-line to rust-mode-indent-line.

If you're suggesting to turn it into one massive function, I think that's unwise.  There is very little overlap in logic/functionality between those two functions: rust-doc-indent-line is concerned with *rustdoc* syntax, not with *rust* syntax.  It's almost completely orthogonal to the programming language.

psibi:
@jcs090218 Do you agree with the comments of @bemoody  ?

@bemoody Thanks for the PR, LGTM from a high level. Is it possible to have a test for this functionality ?

jcs090218:
I suggest renaming rust-mode-indent-line to rust-mode--indent-line (or name anything logical) and creating a new rust-mode-indent-line to call the other two functions, etc.

I think it doesn't really matter since it requires little to no refactoring right now. 🤔

bemoody:
Sure, sounds good!  I haven't worked with ert before, but I think this should do the trick.

psibi (approved):

PR #567, opened 2025-04-04
Upgrade rustfmt edition to 2024

eval-exec:
Fix https://github.com/rust-lang/rust-mode/issues/566

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Update rustfmt's defaults to use 2024 edition ([#566](https://github.com/rust-lang/rust-mode/issues/509)).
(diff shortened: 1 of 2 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

apiraino:
For the record, this change will be neutralized in #571 

I reported in #570 that `rust-rustfmt-switches` was shadowing the `rustfmt.toml`, ignoring any user configuration.

eval-exec:
> For the record, this change will be neutralized in #571 
> 
> I reported in #570 that `rust-rustfmt-switches` was shadowing the `rustfmt.toml`, ignoring any user configuration.

Agree. I think we should set rustfmt-switches to nil by default.

PR #569, opened 2025-06-08
Add compilation regexp to match ‘dbg!’ output

hron:
While debugging tests it’s annoying when ‘compilation-mode’ matches ‘dbg!’ output incorrectly. Currently, rust-mode’s regexps assign it as an error and incorrectly include ‘[’. The proposed change adds a regexp to match such things and assign them info level.

image


hron:
@psibi sure thing! I have added the test for the regexps and did a bit of renaming to match the current variable names. Let me know if you spot anything else. Thanks!

psibi (approved):

psibi:
Thank you!

PR #576, opened 2025-09-24
Don't hang when the beginning of the error can't be found

dabrahams:
Tools that interfere with rustc output can cause the line to be prefixed (see https://github.com/DioxusLabs/dioxus/issues/4547), which can cause the line never to be found, and the loop to continue until the user presses C-g


psibi (commented):
Overall, this looks good to me. Can you add a test for this?

jcs090218 (approved):

Issue #579, opened 2025-12-04
Support Running Tests With Output


Cyncrovee:
Currently, when you run the `rust-test` function it's pretty much equivalent to running `cargo test` in the terminal, and this is very useful! 

However sometimes you need the print output from tests, which in the terminal you can get by running `cargo test -- --show-output`. As far as I can tell, `rust-mode` does not support this functionality out of the box at the moment- so it would be nice to see it added. Maybe with a function called `rust-test-with-output` or something similar.

PR #580, opened 2025-12-05
Add Support for Testing with Output

Cyncrovee:
This PR solves the issue I raised in #579 by adding a new `rust-test-output` function.


Cyncrovee:
Updated with the suggested change. Feel free to merge unless you want more changes, in which case let me know and I'll take a look.

Cyncrovee:
Okay, this should be GTG now. Managed to create two tests, not sure if two is necessary but if not I can remove one.

psibi (approved):

PR #585, opened 2026-04-13
Add keyword `raw`

skr4n:
Append the keywords list to Include the raw borrow operator, stabilized in Rust 2024.

https://doc.rust-lang.org/reference/expressions/operator-expr.html#raw-borrow-operators


jcs090218 (approved):

psibi (approved):
Thanks!

PuercoPop:
I'm confused about the choice of reference, which describes operator expressions not keywords. The [keyword reference](https://doc.rust-lang.org/reference/keywords.html#weak-keywords) lists it as a "Weak Keyword". Which means it is only a  keyword in certain contexts. If we see the list of weak keywords, except for `static` the rest are not listed as keywords.

And using font-lock for context sensitive highlighting seems like the wrong tool for the job.

jcs090218:
Yeah, you are right. Probably need to take care in special cases like other weak keywords.

skr4n:
My bad, thanks for pointing this out.
Fixed this in #586

PR #586, opened 2026-04-14
Add raw borrow operator as a contextual keyword

skr4n:
The `raw` borrow operator is a contextual keyword.
Highlight it only in the context of borrows.
```rust
let mut x = 5;

// highlight "raw" here
let a = &raw const x;
let b = &raw mut x;

// do not highlight "raw" here
let raw = &x as *const _;
let raw_mut = &x as *mut _;
```

Reverts #585


jcs090218 (approved):

psibi (approved):
Thanks for the quick fix!

PR #587, opened 2026-05-12
imenu: also support const and static items

legendre6891:
This makes imenu also list const and static declarations, e.g. https://doc.rust-lang.org/rust-by-example/custom_types/constants.html


psibi (commented):
Seems like this is causing some test failures.

PR #588, opened 2026-05-19
Handle thread IDs in panic output

roife:
Rust 1.91 includes the OS thread ID in panic messages, for example:

```text
thread 'aaa::bbb' (19178688) panicked at crates/a/src/b.rs:13:9:
```

The existing panic regexp only handled the older form without the thread ID:

```text
thread 'aaa::bbb' panicked at crates/a/src/b.rs:13:9:
```

This updates `rustc-panics-compilation-regexps` to accept the optional thread ID.

**Note:** This is a backward-compatible regexp change. It keeps matching the pre-1.91 panic format and adds support for the new format with an optional thread ID.

**Ref:** https://github.com/rust-lang/rust/pull/115746.

### DIFF
--- a/rust-compile.el
+++ b/rust-compile.el
@@ -41,9 +41,10 @@ See `compilation-error-regexp-alist' for help on their format.")
 See `compilation-error-regexp-alist' for help on their format.")
 
 (defvar rustc-panics-compilation-regexps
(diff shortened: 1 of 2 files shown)


psibi (approved):
Thank you!

PR #589, opened 2026-05-31
Fix byte-compile warnings and correct minimum Emacs version

dannywillems:
This PR resolves the byte-compile warnings emitted by rust-mode and
corrects the declared minimum Emacs version to match the features the
package actually uses. All changes are behaviour-preserving.

### Obsolete `when-let` macro (rust-prog-mode.el)

`when-let` is obsolete as of Emacs 31.1 and emits a byte-compile
warning. Because the package supports older Emacs versions, `when-let*`
(introduced in Emacs 26.1) is not a drop-in replacement at the lowest
supported version. The three usages are rewritten using `let` plus
`when`, available on all supported versions, preserving the original
short-circuiting evaluation order.

### Incorrect minimum Emacs version

rust-cargo.el calls `file-local-name`, introduced in Emacs 26.1, so the
package never ran on the declared minimum of Emacs 25.1. The
Package-Requires header is updated to `((emacs "26.1"))`, matching the
oldest version already exercised by the test matrix.

rust-cargo.el also uses `string-trim` from subr-x, which was only
required at compile time via rust-mode.el. On Emacs versions before 28
(where subr-x is not preloaded) this leaves the symbol undefined at
runtime and unknown to the byte-compiler. A runtime `(require 'subr-x)`
is added to rust-cargo.el.

## CI

A new `byte-compile.yml` workflow byte-compiles the package with
`byte-compile-error-on-warn` enabled, on the minimum (26.1), the current
stable (30.1), and the development snapshot, so new warnings fail CI.
The treesitter integration file is only compiled on Emacs 29.1 and
later, matching its runtime guard. GitHub Actions stay current via the
existing dependabot configuration.

## Residual warnings

None. The package byte-compiles cleanly with warnings as errors on
Emacs 26.1, 30.1, and snapshot.

### DIFF
--- a/.github/workflows/byte-compile.yml
+++ b/.github/workflows/byte-compile.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         emacs-version:
-          - "25.1"
+          - "26.1"
           - "30.1"
           - snapshot
     steps:
--- a/.github/workflows/byte-compile.yml
+++ b/.github/workflows/byte-compile.yml
@@ -1,0 +1,44 @@
+name: Byte-compile
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  byte-compile:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        emacs-version:
+          - "25.1"
+          - "30.1"
+          - snapshot
+    steps:
+      - uses: actions/checkout@v5
+      - uses: purcell/setup-emacs@master
+        with:
+          version: ${{ matrix.emacs-version }}
+      - name: Byte-compile with warnings as errors
+        run: |
+          emacs -Q --batch -L . \
+            --eval '(setq byte-compile-error-on-warn t)' \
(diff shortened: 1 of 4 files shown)


dannywillems:
This is an automated effort to help maintaining packages in the Emacs community. See https://x.com/dwillems42/status/2060720730338185699 and https://dannywillems.github.io/emacs-package-maintenance/

PuercoPop:
> when-let* (introduced in Emacs 26.1) is not a drop-in replacement at the lowest supported version.

Technically true as it uses let* instead of let. In practice it is a drop-in replacement in most of the cases. In the 3 instances in covered in this PR, when-let can be replaced with when-let*. Note that there is no Emacs 31, it hasn't been released yet. It will be soon as the release branch has been cut. So the warning happens only in unreleased versions of Emacs (ej. when running emacs from HEAD).

Similarly for the CI change. The tool rust-mode uses for GitHub actions already supports treating warnings as errors (See https://emacs-eask.github.io/Getting-Started/Commands-and-options/#---strict). Instead of adding a new workflow that doesn't follow the same conventions of the other workflow we should update the existing workflow to compile with byte-compile-error-on-warn set to t.

Note that emacs 26 was released in 2018. Not even debian bullseye has a version that old.https://packages.debian.org/search?keywords=emacs, if something doesn't work with emacs 26.1 it might be better to drop it from the supported versions.

Don't know the maintainers feel about LLM-generated PRs, but it feels incredibly disrespectful to other people to other people to not even review the output of the LLMs for basic sanity checks before opening a PR.

